PR #3, opened 2017-06-21
Feature/iat 489

bdrainer:
I will create a PR for IMS as well.  There are dependent changes.  I think we are in a place where changes to the common lib won't break IMS.  And certainly we'll strive to make any common lib changes backward compatible.

The main change here is to delete the local repo at the end of every request.  This means every request starts by cloning the remote repository to the local file system.  

I will likely continue working on this branch to add back the unit tests I removed.  I did so to get this to Ray for the IAT integration for simplifying the API. 

Another quick note and I will put this on the README is when creating a new branch of the common library, it is important to increment the snapshot version.  If not  the CI build will push it to Artifactory where potentially IMS will pick it up and fail if there incompatible changes made to the common lib.  I think this is what happened last night when IMS's build broke.  Ray merged to develop, develop built on TeacmCit and picked up the most recent common lib snapshot which was not compatible.


src/main/java/org/opentestsystem/ap/common/client/GitClient.java
```diff
@@ -77,62 +78,138 @@
...
-        this.parsedUsername = userInfo.parseUsername();
-        this.repository = repository;
+        this.itemId = itemId;
         this.saaifAssembler = saaifAssembler;
+
         this.credentialsProvider = new UsernamePasswordCredentialsProvider(
             itemBankProperties.getUser(), itemBankProperties.getPassword());
+
+        this.localBaseDir = itemBankProperties.getLocalBaseDir();
+        this.randomFolderName = RandomStringUtils.randomAlphabetic(itemBankProperties.getRandomRepoFolderCharLength());
```
rmitchell3: Looks good.  I was thinking last night we could use the transaction ID for the random folder name but keeping it completely random keeps it simpler.  No need to use the transaction ID for this.

src/main/java/org/opentestsystem/ap/common/client/GitClient.java
```diff
@@ -77,62 +78,138 @@
...
+     * Adds the SAAIF item file to the repository.  The assumption is the repository does not contain
+     * the item file.  The action is to add the item file for the first time and push it to the remote
+     * repository.  The action is against the scratch pad branch which is assumed not to exist so it is
+     * created.
+     *
+     * @param saaifItem     The item file to add.
+     * @param transactionId The transaction to associate the changes with.
+     * @param commitMessage The commit message to associate with the changes.
+     */
+    public void beginCreateTransaction(final ItemRelease saaifItem, final String transactionId, final String commitMessage) {
```
rmitchell3: I really like the series of method calls here.  The code is very self-documenting.

src/main/java/org/opentestsystem/ap/common/config/ItemBankProperties.java
```diff
@@ -21,14 +21,26 @@
...
+ * default value.
  */
 @Getter
 @Setter
 @Component
 @ConfigurationProperties(prefix = "itembank")
 public class ItemBankProperties {
 
+    // Min value for the Item ID generator
+    private int idMinValue = 1000000000;
```
rmitchell3: Are these constants?  If so should they be final and should the be named in all caps?  Not sure on the convention in Java.

src/test/java/org/opentestsystem/ap/common/client/GitClientTest.java
```diff
@@ -554,19 +575,128 @@ public void test() {
...
+        final CloneCommand cloneCommand = gitClient.cloneCommand();
+        assertThat(cloneCommand).isNotNull();
+    }
+
+    @Test
+    public void itShouldCreateNewRefSpec() {
+        final RefSpec refSpec = gitClient.newRefSpec();
+        assertThat(refSpec).isNotNull();
+    }
+
```
rmitchell3: Nice tests.

rmitchell3:
Looks great.  Only question I had was regarding the naming convention of constants in Java (see the comment).  Very nice changes and clear code.

src/main/java/org/opentestsystem/ap/common/config/ItemBankProperties.java
bdrainer: This class is instantiated when the app loads and the values are set from what is in application.yml (or served up by the config service).  The defaults set in the class mean one does not have to set them in the application.yml unless they want to override the default.

I guess the config property is kind of hidden and only known to those with insight into the ItemBankProperties class.  Maybe it would be better to always define them in the application.yml?

rmitchell3: Gotcha.  The way it is sounds good then.

PR #4, opened 2017-06-22
Feature/iat 268

bdrainer:
The new method on the repository has been created.  

@alex-ponce  - I updated the snapshot version of the common lib to 0.1.3-SNAPSHOT.  Technically you shouldn't have to build it yourself and publish it to local maven.  The reason is the CI build pushes the snapshot to Artifactory.  The trick though is your IRS build.gradle needs to be configured to look where the snapshots are located.  IMS already has this configured.  It is under the 'artifactory' section.  This is the main part to be aware of.

```
    resolve {
        repository {
            repoKey = 'libs-releases'
            maven	= true
        }
        repository {
            repoKey = 'libs-snapshots'
            maven	= true
        }
    }
```


bdrainer:
@rmitchell3  - I am not sure the default branch of 'develop' is taking.  I just noticed it was using 'master' still as the base branch.

PR #5, opened 2017-06-24
Feature/iat 481

bdrainer:
The main change is ItemRepository.lockItemForEdit.  If the requesting user is not able to lock the item we now look up who created the lock.  This should fix the issue where the name is showing as null in IAT.    Previously when the user couldn't get the lock we would look up the who created the scratch pad, but the scratch pad may not have been created yet, so that is why null was being returned.

Another nice change was handling errors and properly cleaning up.  This is important because if, for example, begin edit transaction locks the item but fails we don't want to leave the item locked.  We now catch the error and clean up.


rmitchell3:
Looks good.

PR #13, opened 2017-07-08
Persist MC in SAAIF format

alex-ponce:
* Created McItem and ItemOption model classes
* Implemented mapping to and from IMS and SAAIF models
* Implemented optionlist and option JAXB configuration in ItemMapper and ObjectFactory classes
* Validated all workflow operations work via IMS REST calls:
** Begin Transactions
** Create Transactions
** Begin Edit
** Save
** Commit
** Rollback
** Get Item
** Get Item History
** Get Item Diff

_Save Operation Payload_

**PATCH** http://localhost:8081/api/v1/items/2036135617/transactions/1806358003
```json
{
	"message": "saving MC transaction",
	"item" : {
		"type": "mc",
		"prompt": "Saving MC via REST endpoint",
		"options": [
			{
				"text": "Response A", 
				"correctAnswer": "false"
			},
			{
				"text": "Response B", 
				"correctAnswer": "false"
			},
			{
				"text": "Response C", 
				"correctAnswer": "false"
			},
			{
				"text": "Response D", 
				"correctAnswer": "true"
			}
		],
		"stimulusId": "1171900071"
	}
}
```


src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -257,23 +249,54 @@ public static void addUpdateOrRemoveAttribute(final ItemRelease.Item.Attriblist
...
         }
         return Collections.emptyList();
     }
 
+    public static List<ItemOption> findOptions(final ItemRelease.Item.Content saaifContent,
+                                               final ItemRelease.Item.Attriblist saaifAttribs) {
+
+        ItemRelease.Item.Attriblist.Attrib keyAttrib =
+                saaifAttribs.getAttrib().stream().filter(attrib ->
+                        ItemConstants.ItemAttribute.ItemAnswerKey.getAttId().equals(attrib.getAttid())).
```
bdrainer: If you wanted to shorten the verbosity referencing the num you could use a static import reducing it to ItemAnswerKey.getAttId()

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -257,23 +249,54 @@ public static void addUpdateOrRemoveAttribute(final ItemRelease.Item.Attriblist
...
+                        collect(Collectors.toList()).get(0);
```
bdrainer: I'm more thinking out loud here..., we are guaranteed to have ItemAnswerKey in the attribute list because when the item is first created it is added.  Once added it is never removed so should  always be there.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -257,23 +249,54 @@ public static void addUpdateOrRemoveAttribute(final ItemRelease.Item.Attriblist
...
+        if (keyAttrib != null) {
+            attribAnswer = keyAttrib.getVal();
+        }
+
+        List<ItemOption> options = new ArrayList<>();
+        if (saaifContent != null) {
+            final ItemRelease.Item.Content.Optionlist saaifOptionlist = saaifContent.getOptionlist();
+            if (saaifOptionlist != null) {
+                for (int index = 0; index < saaifOptionlist.getOption().size(); ++index) {
+                    ItemOption option = new ItemOption();
```
bdrainer: I was using a builder on our entities.  I installed an IntelliJ plugin to do it so it is easy generator the build class.  I like the builder but realize I'm defeating the purpose by having setters on the our entities.  My thinking was SaItem, WerItem, StimItem, etc. have getters and no setters, so they are immutable.  But they do have setters which isn't what I wanted.  Let me know your thoughts.  Regardless you should check out the plugin - https://plugins.jetbrains.com/plugin/6585-builder-generator  -  it has a short cut key of ALT+SHIFT+B

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -358,4 +381,74 @@ public static void mapToSaaifExemplarResponses(final List<String> exemplarRespon
...
+            saaifContent.setOptionlist(mapOptions(options));
+        }
+    }
+
+    public static ItemRelease.Item.Content.Optionlist mapOptions(final List<ItemOption> options) {
+
+        final ItemRelease.Item.Content.Optionlist saaifOptionslist = OBJECT_FACTORY
+                .createItemreleaseItemContentOptionlist();
+
+        if (saaifOptionslist != null) {
```
bdrainer: check if the call before this on OBJECT_FACTORY will ever return null, if not you could skip this null check on saaifOptionslist.  I doubt it returns null.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -358,4 +381,74 @@ public static void mapToSaaifExemplarResponses(final List<String> exemplarRespon
...
+                        EMPTY, EMPTY);
+
+                final String option = options.get(index).getText();
+                if (StringUtils.isNotBlank(option)) {
+                    saaifOption.setName("Option " + getCapitalLetterForIndex(index));
+                    saaifOption.setVal(option);
+                    saaifOption.setFeedback(EMPTY);
+                }
+
+                saaifOptionslist.getOption().add(saaifOption);
```
bdrainer: A Java 8 way to do this is create a static method that takes an ItemOption and returns an Optionlist.Option.  I see you have a mapToSaaifOption taking three parameters.   The one I'm suggesting takes a single parameter of type ItemOption for example 

"public Optionlist.Option mapToSaaifOption(ItemOption)"  This would do most of what your doing now in the for loop.

And instead of the for loop, stream over the list of options something like 

List<Optionlist.Option> saaifOptions = options.stream().map(ItemMapper::mapToSaaifOption).collect(Collectors.toList());

Then call saaifOptionlist.getOption().addAll(saaifOptions).  

I'm still getting used to Java 8 streams but they are pretty cool for some things.

bdrainer: I see there is a little more two it with mapping to a Optionlist.Option.  You need to get the capital letter using the index.  That complicates it and the way I suggested would need to be reconsidered with some other type of streaming or just use the good old for loop.  I would suggestion a mapping from ItemOption to Optionlist.Option.  Maybe that mapping method takes an argument "int index" so the it has the index to get the capital letter.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -358,4 +381,74 @@ public static void mapToSaaifExemplarResponses(final List<String> exemplarRespon
...
+    }
+
+    /**
+     * Return Capital letter that corresponds to the order in which options are processed
+     *
+     * @param index
+     * @return
+     */
+    private static String getCapitalLetterForIndex(int index) {
+        return Character.toString((char) (index + 65));
```
bdrainer: I'm guessing adding 65 to the int index is the capital letter where 0 + 65 must start at A.  It would be more clear to make 65 a static final maybe "static final int CAPITAL_A = 65;"

src/test/java/org/opentestsystem/ap/common/it/IntegrationTestUtil.java
```diff
@@ -51,7 +51,7 @@ public void init() {
         props.setApiVersion("/api/v4");
         props.setBankKey("187");
         props.setGroup("iat-development");
-        props.setLocalBaseDir("/Users/brettdrainer/ItemBank");
+        props.setLocalBaseDir(getenv("HOME") + "/ItemBankCommon");
```
bdrainer: nice

PR #14, opened 2017-07-11
Feature/iat 224

bdrainer:
Added saving the metadata.xml to the item bank with the new element WorkflowStatusCode.


src/main/java/org/opentestsystem/ap/common/saaif/SaaifMetadataAssembler.java
```diff
@@ -0,0 +1,153 @@
...
+    public void init() {
+        try(final InputStream is = this.getClass().getClassLoader().getResourceAsStream(SCHEMA_FILE)) {
+            schema = SCHEMA_FACTORY.newSchema(new StreamSource(is));
+            jaxbContext = newInstance(SmarterAppMetadata.class);
+        } catch (Exception e) {
+            throw new SystemException("Error instantiating Schema or JAXBContext", e);
+        }
+    }
+
+    // ------------------------------------------------------------------------
```
rmitchell3: What are these comment bars?

src/test/java/org/opentestsystem/ap/common/it/IntegrationTestUtil.java
```diff
@@ -39,7 +42,10 @@
     public void init() {
         itemFactory = new ItemFactory();
 
-        saaifAssembler = new SaaifAssembler();
+        metadataAssembler = new SaaifMetadataAssembler();
+        metadataAssembler.init();
```
rmitchell3: Why call init after constructing?  Does the constructor not fully init?

rmitchell3:
Looks great!

src/main/java/org/opentestsystem/ap/common/saaif/SaaifMetadataAssembler.java
bdrainer: I might have a bit of OCD on this.  I tend to use these to section off things.  I almost always have my classes follow a pattern where the top portion is statics finals, then statics, then properties, then constructors where I'll follow the constructor (and sometimes an @PostConstruct) with  divider.  That would be my class header section.  Then I have public methods followed by divider.  Then protected and private methods followed by divider.  Finally I have getters/setters at the very bottom.  I might have gotten used to the visual cue of the divider as well like when I'm scrolling the file.   I can take these out as no one else does it and it doesn't add value.  It is likely IMS and ap-common have these throughout.

rmitchell3: No worries.  No need to remove.  I figured it was something like that.  You'd really like regions in C#.

PR #15, opened 2017-07-11
Fixed IAT-556; Completed NormalItem refactoring

alex-ponce:
* Changed McItem.correctAnswer to McItem.isCorrectAnswer
* On McItem type Items, cleared itm_att_Answer Key value upon item creation and when a correct answer is never passed in the optionslist
* Finalizes NormalItem refactoring. Moved ```private List<String> exemplarResponses;``` into SaItem.java and WerItem.java


src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -58,6 +58,7 @@
 public class ItemMapper {
 
     private static final ObjectFactory OBJECT_FACTORY = new ObjectFactory();
+    private static final int CAPITAL_A = 65;
```
rmitchell3: You could eliminate this constant and just convert 'A' to it's numeric value where you need it as shown here:  https://stackoverflow.com/questions/16458564/convert-character-to-ascii-numeric-value-in-java.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -296,11 +297,11 @@ public static void addUpdateOrRemoveAttribute(final ItemRelease.Item.Attriblist
                 for (int index = 0; index < saaifOptionlist.getOption().size(); ++index) {
                     ItemOption option = new ItemOption();
                     option.setText(saaifOptionlist.getOption().get(index).getVal());
-                    boolean correctAnswer = false;
+                    boolean isCorrectAnswer = false;
```
rmitchell3: This variable declaration plus the if could be combined into one statement:  `boolean isCorrectAnswer = attribuAnswer.equals(getCapitalLetterForIndex(index));`

src/main/java/org/opentestsystem/ap/common/model/McItem.java
```diff
@@ -28,7 +29,7 @@
 
     public static String RESPONSE_TYPE = "Vertical";
 
-    private static List<ItemOption> options;
+    private List<ItemOption> options;
```
rmitchell3: Good to get rid of static.

src/main/java/org/opentestsystem/ap/common/model/McItem.java
```diff
@@ -104,10 +113,10 @@ public McItemBuilder prompt(String prompt) {
             return this;
         }
 
-        public McItemBuilder options(List<ItemOption> options) {
-            setOptions(options);
-            return this;
-        }
+//        public McItemBuilder options(List<ItemOption> options) {
```
rmitchell3: Should remove commented-out code or add a comment explaining why it's still here.

rmitchell3:
Looks good.  Just a few points of feedback (see comments).

PR #16, opened 2017-07-12
Modified ItemOption getters and setters to allow isCorrectAnswer Json…

alex-ponce:
* Modified ItemOption getters and setters to allow isCorrectAnswer Json property
* Removed commented out code
* Updated unit test


src/main/java/org/opentestsystem/ap/common/model/ItemOption.java
```diff
@@ -1,14 +1,31 @@
...
  * Created by alexponce on 7/6/17.
  */
-@Data
 public class ItemOption {
 
     private String text;
 
     private boolean isCorrectAnswer;
+
+    @JsonProperty("text")
```
bdrainer: Interesting we needed to add these properties and remove Lombok.  We haven't had to do this with other common models.  I would expect the 'text' property to convert to json as the property name 'text' without needing the annotation.  

Regarding the boolean, I think in the Java world it is common to name the property 'correctAnswer' and not 'isCorrectAnswer'.  The setter would be setCorrectAnswer and the getter would be isCorrectAnswer.  I wonder if naming the java property 'isCorrectAnswer' is what caused the original problem.  I suspect Lombok was naming the getter isIsCorrectAnswer and the setter setIsCorrectAnswer.  Possibly that threw the json serialization off.

PR #21, opened 2017-07-15
Added persistence to MS Item type

alex-ponce:
* Added persistence to MS Item Type
* In ```ItemMapper.java```, refactored ```mapToSaaifAnswerKey``` and ```findOptions``` to support single letter and comma delimited values


src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -34,22 +26,20 @@
...
-import static org.opentestsystem.ap.common.model.ItemConstants.EXEMPLAR_RESPONSE;
-import static org.opentestsystem.ap.common.model.ItemConstants.EXEMPLAR_SCOREPOINT;
-import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.ItemAnswerKey;
-import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.ItemFormat;
-import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.ItemId;
-import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.ItemPageLayout;
-import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.ItemResponseType;
+import static org.apache.commons.lang3.StringUtils.*;
+import static org.opentestsystem.ap.common.model.ItemConstants.*;
+import static org.opentestsystem.ap.common.model.ItemConstants.ItemAttribute.*;
```
bdrainer: The convention is not to use splats (i.e. the '*') in imports.  You can set IntelliJ so there are no splats when doing imports and static imports.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -436,23 +425,47 @@ public static void mapToSaaifOptions(final List<ItemOption> options,
...
+            }
         }
+
         ItemRelease.Item.Attriblist.Attrib answerKeyAttrib =
                 saaifAttribs.getAttrib().stream().filter(attrib ->
                         ItemConstants.ItemAttribute.ItemAnswerKey.getAttId().equals(attrib.getAttid())).
                         collect(Collectors.toList()).get(0);
         if (answerKeyAttrib != null) {
-            answerKeyAttrib.setVal(answerValue);
+            answerKeyAttrib.setVal(answerValue.toString());
```
bdrainer: It looks like answerValue is already a string so you don't need to do a toString() on it.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -463,6 +476,6 @@ public static void mapToSaaifAnswerKey(final List<ItemOption> options,
      * @return
      */
     private static String getCapitalLetterForIndex(int index) {
-        return Character.toString((char) (index + CAPITAL_A));
+        return Character.toString((char) (index + (int) 'A'));
```
bdrainer: Nice

src/main/java/org/opentestsystem/ap/common/model/McItem.java
```diff
@@ -70,7 +70,7 @@ public ItemRelease toSaaif(ItemRelease saaifItemRelease) {
         final ItemRelease.Item.Content englishContent = findEnglishContent(saaifItem);
 
         mapToSaaifOptions(getOptions(), englishContent);
-        mapToSaaifAnswerKey(getOptions(), saaifItem.getAttriblist());
+        mapToSaaifAnswerKey(ItemConstants.ItemType.TYPE_MC, getOptions(), saaifItem.getAttriblist());
```
bdrainer: You could use getType() here instead of ItemType.TYPE_MC.  Same for MS below.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
alex-ponce: I ran the IntelliJ optimize imports function. Will configure it not to do this again

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
alex-ponce: I initially was obtaining the string from the string builder

PR #25, opened 2017-07-21
IAT-569:  Small refactoring in ItemMapper.  Minor clean up like using…

bdrainer:
Small refactoring


src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -451,31 +450,27 @@ public static void mapToSaaifOptions(final List<ItemOption> options,
...
         if (options != null) {
             StringBuilder sb = new StringBuilder();
             for (int index = 0; index < options.size(); ++index) {
                 if (options.get(index).isCorrectAnswer()) {
-                    if (itemType.equals(TYPE_MC)) {
+                    if (TYPE_MC.equals(itemType)) {
                         // ItemType.TYPE_MC is a single capital letter
                         sb.append(getCapitalLetterForIndex(index));
-                    } else if (itemType.equals(TYPE_MS)) {
+                        break;
```
bdrainer: I put a break here.  Let me know if there are any issues with this.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -451,31 +450,27 @@ public static void mapToSaaifOptions(final List<ItemOption> options,
...
                                            final List<ItemOption> options,
                                            final ItemRelease.Item.Attriblist saaifAttribs) {
-        String answerValue = "";
+        String answerValue = EMPTY;
         if (options != null) {
             StringBuilder sb = new StringBuilder();
             for (int index = 0; index < options.size(); ++index) {
                 if (options.get(index).isCorrectAnswer()) {
-                    if (itemType.equals(TYPE_MC)) {
+                    if (TYPE_MC.equals(itemType)) {
```
bdrainer: I like having a constant be what is compared in the case itemType is null.  It avoids a NPE.

Would it be better if we did an equalsIgnoreCase instead of equals?  It would be more forgiving in the event existing data is not as proper as we hope.

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -451,31 +450,27 @@ public static void mapToSaaifOptions(final List<ItemOption> options,
...
+            answerValue = TYPE_MS.equals(itemType) ? sb.toString().replaceAll(",$", EMPTY) : sb.toString();
         }
 
-        ItemRelease.Item.Attriblist.Attrib answerKeyAttrib =
-                saaifAttribs.getAttrib().stream().filter(attrib ->
-                        ItemAnswerKey.getAttId().equals(attrib.getAttid())).
-                        collect(Collectors.toList()).get(0);
+        final ItemRelease.Item.Attriblist.Attrib answerKeyAttrib = findSaaifAttribute(saaifAttribs, ItemAnswerKey);
+
         if (answerKeyAttrib != null) {
```
bdrainer: if answerKeyAttrib happens to be null should be go ahead an create and add it to the map.  It shouldn't be null as we make sure to create it when creating a new MS item, but going back to existing data and just trying to be as forgiving as we can.  Not a big deal, but just some thoughts.

PR #29, opened 2017-07-27
Feature/iat 317

bdrainer:
A handful of new classes to represent the ItemBody and its contents.  There has to be a better way, most specifically when it comes to the HTML tags like <span> and <p>.  We aren't using <span> but it was in the example EBSRs.  Either way we can cross that bridge when we get there.    For now, all the "prompts" in the EBSR have a <p> tag with the prompt value in them.  I wonder if that was even necessary but again just copying what I was seeing in the examples.


src/main/java/org/opentestsystem/ap/common/mapper/EbsrMapper.java
```diff
@@ -0,0 +1,485 @@
+package org.opentestsystem.ap.common.mapper;
```
alex-ponce: First item type that requires it's own mapper

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -133,24 +140,29 @@ public static ItemRelease newItemRelease(ItemRelease.Item item) {
...
+                                                                        final String itemFormat,
+                                                                        final String itemPageLayout,
+                                                                        final String itemResponseType) {
         final ItemRelease.Item.Attriblist attriblist = OBJECT_FACTORY.createItemreleaseItemAttriblist();
-        attriblist.getAttrib().add(newSaaifAttribute(ItemId, itemId));
-        attriblist.getAttrib().add(newSaaifAttribute(ItemFormat, itemFormat.toUpperCase()));
-        attriblist.getAttrib().add(newSaaifAttribute(ItemAnswerKey, itemFormat.toUpperCase()));
-        attriblist.getAttrib().add(newSaaifAttribute(ItemPageLayout, itemPageLayout));
-        attriblist.getAttrib().add(newSaaifAttribute(ItemResponseType, itemResponseType));
+        addSaaifAttribute(attriblist, ItemId, itemId);
```
alex-ponce: Good refactoring. Less verbose

src/main/java/org/opentestsystem/ap/common/mapper/EbsrMapper.java
bdrainer: Yes.  It uses the new ItemBodyAssembler which is what makes it the most different.  ItemBody and everything inside it like ChoiceInteraction and SimpleChoice.

src/main/java/org/opentestsystem/ap/common/model/EbsrItem.java
```diff
@@ -0,0 +1,150 @@
...
+        final ItemRelease.Item.Content.Qti qti = newQtiForItemBody(itemBodyXMLString);
+
+        final ItemRelease.Item.Content englishContent = findEnglishContent(saaifItem);
+        englishContent.setQti(qti);
+
+        return itemRelease;
+    }
+
+    @Override
+    public ItemRelease toSaaif(ItemRelease saaifItemRelease) {
```
alex-ponce: Very clean toSaaif and fromSaaif methods!

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -164,17 +176,34 @@ public static void addUpdateOrRemoveAttribute(final ItemRelease.Item.Attriblist
...
 
     /**
+     * Finds the attribute first to determine if it exists.  If it exists its value is set to 'value'.  If it does not
+     * exist then it is added to the list with the value set to 'value'.
+     *
+     * @param attriblist The list to add or update.
+     * @param attribute  The attribute to add or update.
+     * @param value      The value to set on the item being added or updated.
+     */
+    public static void addOrUpdateAttribute(final ItemRelease.Item.Attriblist attriblist,
```
alex-ponce: I like how you used this method in the item implementation

src/test/java/org/opentestsystem/ap/common/saaif/ItemBodyAssemblerTest.java
```diff
@@ -0,0 +1,123 @@
...
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class ItemBodyAssemblerTest {
+
+    private final ObjectFactory of = new ObjectFactory();
+
+    private ItemBodyAssembler assembler;
+
+    private final String xmlString = "<?xml version=\"1.0\" encoding=\"UTF-8\" standalone=\"yes\"?><itemBody><p>The " +
+        "following question has two parts.  First answer Part A then answer part B.</p><choiceInteraction " +
```
alex-ponce: Good test sample data. You made them hard to pick

PR #32, opened 2017-07-28
IAT-45:  Added a finder for getting the item history in CSV format.

bdrainer:
Added a CSV mapper for converting a Pojo annotated with Jackson annotations to a CSV string.

There were a couple other cleanup items and some reformatting which explains some of the other file changes in the PR.


src/main/java/org/opentestsystem/ap/common/model/ItemHistoryCsv.java
```diff
@@ -0,0 +1,39 @@
...
+ * The details of a commit on an item in the itembank.  The commit details are from the master branch only. The
+ * intention of this model is to be rendered as CSV.
+ */
+@Data
+@JsonPropertyOrder({"user", "time", "message"})
+public class ItemHistoryCsv {
+
+    private String user;
+
+    @JsonFormat(pattern = "MM/dd/yyyy, hh:mm:ss aaa", timezone = "PST")
```
alex-ponce: This is cool

bdrainer: I thought so too, but the timezone is assuming too much.  I changed it to be formatted to the timezone GitLab returns by formatting it manually and setting the property as a string.  

If we don't put timezone the formatting works but the value is in UTC format.  It didn't look right to see 9:30pm when I created the item at 2:30pm.

PR #34, opened 2017-08-08
Feature/iat 585

bdrainer:
GitlabClient has a new method that gets the history and filters it.  A property has been added called 'historyExclusionPhrase'.  Any commit message containing that phrase is filtered out.

The ItemRepository uses the new method to get the filtered history.  

Calling a different GitLab API to get the item diff.  It uses a range where two commit IDs are required.  This changed the model as the API response is different.


src/main/java/org/opentestsystem/ap/common/client/GitlabClient.java
```diff
@@ -133,8 +133,23 @@ public ItemAuthor lookupItemAuthor(final Project project) {
...
+     * properties 'historyExclusionPhrase'.  The history of an item is the master repository's commit history.
+     *
+     * @param itemId The unique item identifier.
+     * @return An empty list if not history is found or a list containing the commit history of the item.
+     */
+    public List<Commit> lookupItemHistoryFiltered(final String itemId) {
+        log.debug("Look up item history filtered for item {}", itemId);
+        final List<Commit> historyList = lookupItemHistory(itemId);
+        return historyList.stream().filter(c ->
+            !containsIgnoreCase(c.getMessage(), gitlabProperties.getHistoryExclusionPhrase()))
```
alex-ponce: Super clean way of filtering

PR #35, opened 2017-08-13
Feature/iat 66

bdrainer:
The core code to add and remove braille attachments.  

I see there is redundancy between ItemMapper and StimMapper. The difference being the object hierarchy.   One starting at Item and one starting at Passage.  The technical debt cleanup task this sprint, I'm going to look into reducing the code repetition.  

The saaif attachment element has an 'id' attribute.  I'm using that as a timestamp to store when the file was created/updated.  When cloning the repo locally and asking for the file lastModified time it is always the time it was cloned locally.  Another option would be adding metadata elements to track when attached files are updated.  I'd likely mirror the structure in the item file so an attachmentlist element holding a list of attachment elements.  The attachment element could have an attribute for modified date.  Even better would be modifying the saaif spec so the item xml had an attribute on attachment for this value.  Thoughts?


alex-ponce:
Brett. I think using the attachment id to store the timestamp is a very good solution for now. There is very slow risk of duplication and the current actions on an attachment (download and delete) use the filename as an Id. Adding the files attributes to the metadata sounds like the most comprehensive solution with the current saaif spec limitations, however given the current requirements would seem a bit overkill. Modifying the saaif spec would certainly make more sense for a data integrity perspective.

PR #37, opened 2017-08-24
Feature/iat 39

bdrainer:
I tested this one pretty good, but have not tested the TI item with the UI.  I would like to merge this so I can start IAT-666.  If there are TI implementation fixes needed they should be isolated to just the TI code.  I'm thinking they'll merge easily into IAT-666 but starting IAT-666 with what we have for IAT-39 is preferable.


src/main/java/org/opentestsystem/ap/common/client/GitClient.java
```diff
@@ -231,6 +298,37 @@ public void writeItemLocally(final ItemRelease saaifItem, final SmarterAppMetada
         }
     }
 
+    public void attachFile(final Item item, final Attachment attachment, final MultipartFile multipartFile) {
```
alex-ponce: Are attachFile and detachFile still exclusive to Braille files?

src/main/java/org/opentestsystem/ap/common/config/FreemarkerConfig.java
```diff
@@ -0,0 +1,28 @@
+package org.opentestsystem.ap.common.config;
```
alex-ponce: Freemarker. I learn something new everyday. Clever solution to create html markup.

src/main/java/org/opentestsystem/ap/common/client/GitClient.java
bdrainer: This is for more than braille files.  It could be any file and it gets attached to the list of attachments.  IMS is only receiving request for braille but if it exposes APIs for receiving other file types this ap-common method will handle it as well.

src/main/java/org/opentestsystem/ap/common/config/FreemarkerConfig.java
bdrainer: I thought so.  You might have heard of Velocity which I was going to use but apparently that is old and no one is using it anymore.  Freemarker is a good alternative I feel.  Most template engines are for JSP replacements like Thymleaf and Mustache, along with Freemarker.  I found Freemarker a good standalone template engine.

src/main/java/org/opentestsystem/ap/common/model/Braille.java
```diff
@@ -17,5 +18,5 @@
     @JsonProperty("isBrailleProvided")
     private boolean brailleProvided;
 
-    private List<Attachment> attachments;
+    private List<Attachment> attachments = new ArrayList<>(0);
```
alex-ponce: Nothing wrong here. Just curious about the need to initialize the attachment arrays

bdrainer: Without this, the JSON was getting to the UI as null causing the UI to fail.  It was happening when the item was first created.  

You'll see I did this all over the place.  The effect is the json given to the front end has 
"attachments = []" 
instead of 
"attachements = null"

We must have not needed this when we were reading from the XML file and mapping to the common model.

src/main/java/org/opentestsystem/ap/common/saaif/JsonModelAssembler.java
```diff
@@ -0,0 +1,37 @@
...
+import org.opentestsystem.ap.common.model.AbstractItem;
+import org.opentestsystem.ap.common.model.Item;
+import org.opentestsystem.ap.common.util.SystemException;
+import org.springframework.stereotype.Component;
+
+/**
+ * Reads and writes item models to/from JSON format.
+ */
+@Component
+public class JsonModelAssembler {
```
alex-ponce: Such a simple class to do so much. Nice.

src/main/resources/saaif_templates/ti-rubric.ftl
```diff
@@ -0,0 +1,38 @@
+<p style="">Exemplar:</p><p style="">&#xA0;</p>
```
alex-ponce: These files remind me of XSLT

src/main/java/org/opentestsystem/ap/common/config/FreemarkerConfig.java
alex-ponce: Nooo not Velocity! I used it in my previous job and it's terrible and unsupported. Glad you went with Freemarker

src/main/java/org/opentestsystem/ap/common/client/GitClient.java
alex-ponce: Thinking that when we get to integrate images and sound files to content we'll need the ability to upload files without adding a reference to the items attachment list. Not 100% but that's how it looked on a few sample items.

PR #40, opened 2017-09-01
Removed item type prefix from item ID

alex-ponce:
* Removed item type prefix from item ID
* Removed item id replacement in cloneForPreview method


rmitchell3:
Looks good.

PR #41, opened 2017-09-02
Feature/iat 580

bdrainer:
The majority and possibly all the validation work is done in the common lib.  It won't cause any compatibility issues with IMS or IRS.  Only new functionality needed by IVS has been added. 

Let's merge this now so it can get merged into the IAT-666 before IAT-666 gets to different.  I want to avoid a tough merge if we can.


rmitchell3:
This looks great.  Surprisingly few changes needed.

PR #47, opened 2017-09-12
Feature/iat 65

hbgwhite:
ASL tab


src/main/java/org/opentestsystem/ap/common/saaif/metadata/SmarterAppMetadata.java
```diff
@@ -178,6 +183,10 @@
     protected String doesThisItemRequireBraille;
     @XmlElement(name = "HasAllBrailleContentBeenProvided", required = false)
     protected String hasAllBrailleContentBeenProvided;
+    @XmlElement(name = "DoesThisItemRequireAsl", required = false)
+    protected String doesThisItemRequireAsl;
+    @XmlElement(name = "HasAllAslContentBeenProvided", required = false)
+    protected String hasAllAslContentBeenProvided;
```
bdrainer: Well done on the changes in this file.  I wondered if you would find your way into this one.  We originally generated this from an XSD but are hand coding it from here on out.  Everything in here looks good.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -636,7 +639,7 @@ Item saveSectionEdits(final Item changedItem, final Item existingItem, final Str
     void applyAttachment(final Item item, final ItemRelease saaifItem,
                          final String section, final Attachment attachment) {
         log.debug("update attachment item {},  section {}, file {}", item.getId(), section, attachment.getFileName());
-        if (SECTION_CREATE.equalsIgnoreCase(section) || SECTION_BRAILLE.equalsIgnoreCase(section)) {
+        if (SECTION_CREATE.equalsIgnoreCase(section) || SECTION_BRAILLE.equalsIgnoreCase(section)  || SECTION_ASL.equalsIgnoreCase(section)) {
             item.attachBrailleFile(saaifItem, attachment);
```
bdrainer: I think you want an 'else if' separating BRAILLE from ASL.  Notice when it is braille we call item.attachBrailleFile(...).   I'm guessing when it is ASL (or CREATE) we should call item.attachAslFile(...)

If the section is CREATE that means any section can be edited.

src/main/java/org/opentestsystem/ap/common/model/Item.java
```diff
@@ -76,4 +80,9 @@
 
     ItemRelease detachBrailleFile(final ItemRelease saaifItem, final String fileName);
 
+    Item updateItemAsl(final Item existingItem);
+
+    ItemRelease attachAslFile(final ItemRelease saaifItem, final Attachment attachment);
+
+    ItemRelease detachAslFile(final ItemRelease saaifItem, final String fileName);
```
bdrainer: Where does detachAslFile get called?

src/main/java/org/opentestsystem/ap/common/mapper/ItemMapper.java
```diff
@@ -615,6 +617,23 @@ public static void removeBrailleModelfAttachment(final Braille braille, final St
...
+        Optional<Attachment> optional = Optional.empty();
+
+        if (isNotEmpty(attachments)) {
+            optional = attachments.stream().filter(a -> equalsIgnoreCase(a.getFileName(), fileName)).findFirst();
+        }
+
+        if (optional.isPresent()) {
+            Attachment attachment = optional.get();
+            final boolean removed = attachments.remove(attachment);
+            log.info("braille attachment {} was removed = {}", fileName, removed);
```
bdrainer: The log.info references 'braille' but it should be 'asl'.

src/main/java/org/opentestsystem/ap/common/client/GitClient.java
```diff
@@ -350,7 +350,7 @@ public void attachFile(final Item item,
      * @param fileName      The file to delete.
      * @param commitMessage The commit message associated with the change.
      */
-    public void dettachFile(final Item item, final ItemRelease saaifItem,
+    public void detachFile(final Item item, final ItemRelease saaifItem,
```
bdrainer: Good catch.  Thank you.

PR #52, opened 2017-09-13
Feature/iat 67

hbgwhite:
Attachments during create


src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -674,7 +674,7 @@ void applyAttachment(final Item item, final ItemRelease saaifItem,
                          final String section, final Attachment attachment) {
         log.debug("update attachment item {},  section {}, file {}", item.getId(), section, attachment.getFileName());
         if (sectionMatchesFileType(section, attachment)) {
-                item.attachFile(saaifItem, attachment,  getAttachmentFileTypeBySection(section));
+                item.attachFile(saaifItem, attachment,  attachment.getAttachmentFileType());
```
bdrainer: Not a big deal, more of an observation, we don't need the third argument here because the attachment has the file type and we are passing the attachment as the second arg.

PR #57, opened 2017-09-20
Feature/iat 814

hbgwhite:
Restructuring validation object to return additional information


src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -597,7 +647,15 @@ public ItemResponse findItem(final ItemBankUser user, final String itemId) {
...
         final List<ItemHistory> itemHistoryList = commits.stream().map(ModelMapper::mapToEntityItemHistory).collect
-            (toList());
+                (toList());
+        return itemHistoryList;
+    }
+
+    public List<ItemHistory> findFullItemHistory(final String itemId) {
+        log.debug("find full item history for {}", itemId);
+        final List<Commit> commits = gitlabClient.lookupItemHistory(itemId);
+        final List<ItemHistory> itemHistoryList = commits.stream().map(ModelMapper::mapToEntityItemHistory).collect
```
bdrainer: This one is on me, we should rename ModelMapper.mapToEntityItemHistory to mapToItemHistory.  Since moving the method to the ModelMapper it is redundant to use 'Entity'.

hbgwhite: swapped out the method signature to remove entity for this one and for the csv version

src/main/java/org/opentestsystem/ap/common/model/ValidationResults.java
```diff
@@ -0,0 +1,14 @@
+package org.opentestsystem.ap.common.model;
+
+import lombok.Data;
+
+import java.util.List;
+
+@Data
+public class ValidationResults {
+    public boolean areValidationResultsStale;
```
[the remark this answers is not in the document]
hbgwhite: done

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -148,30 +154,74 @@ public void saveValidationResults(final ItemBankSystemUser systemUser,
...
+                .filter(x -> x.getHistoryId().equalsIgnoreCase(commitId))
+                .findFirst();
+        if (result.isPresent()) {
+            return result.get();
+        }
+        throw new ValidationException("Unrecognized commit ID passed to history request " + commitId);
+    }
+
+    private ItemHistory getLatestItemHistory(String itemId, boolean validationOnly) {
+        List<ItemHistory> historyItems = validationOnly ? findFullItemHistory(itemId)
```
bdrainer: What if we made findFullItemHistory always sort from most recent to least.  Actually it does that now by default, the API call must do it (i.e. Gitlab REST API is sorting it for us).  I'm not sure if we should depend on that but at present we are.   But..., let's say findFullItemHistory sorts for us, does that simplify getLastItemHistory?  You could grab the first one, check it it is a validation commit, skip it if necessary.  That removes filtering (kind of), sorting, and reversing from the method.

bdrainer: There is a method on GitlabClient called 'lookupItemHistoryFiltered'.   That will filter out the validation messages.  I like how you are doing it with a regex, probably better than what I did in 'lookupItemHistoryFiltered'.  

Just wanted to throw that method out in case you could use it.  You could add a method to GitLabClient that returns only the validation commits too.   

Also take a look at https://docs.gitlab.com/ce/api/commits.html#get-a-single-commit  - would this be a good method to implement in GitlabClient?  We could look up a commit directly and map it to ItemHistory.

src/main/java/org/opentestsystem/ap/common/mapper/ModelMapper.java
```diff
@@ -30,7 +30,7 @@
      * @param gitlabCommit The Gitlab API model
      * @return The common model.
      */
-    public static ItemHistory mapToEntityItemHistory(final Commit gitlabCommit) {
+    public static ItemHistory mapToItemHistory(final Commit gitlabCommit) {
```
bdrainer: Great.  Thanks for doing that.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
hbgwhite: - I like it better with the list, just because when looking up the validated commit we already have the information we need without an additional call.
- I'll check out the lookupItemHistoryFiltered method and see if I can work it in instead of rolling my own

PR #63, opened 2017-09-27
Feature/iat 33

hbgwhite:
preview


src/main/java/org/opentestsystem/ap/common/mapper/MiMapper.java
```diff
@@ -28,201 +31,113 @@
...
+        List<Table.Column> columns = item
+                .getCore()
+                .getTable()
+                .getColumns();
+        return columns.size() > 1 ? item
+                .getCore()
+                .getTable()
+                .getColumns()
+                .stream()
+                .skip(1)
```
bdrainer: Nice.  "skip" who knew.

gradle.properties
```diff
@@ -2,7 +2,7 @@ description="Micro-service exposing API CRUD operations for assessment items sto
 
 group=org.opentestsystem.ap
 
-version=0.1.28-SNAPSHOT
+version=0.1.29-SNAPSHOT
```
bdrainer: Did you update this?  It should be ok but usually only the release build updates this.  However, I do think there are scenarios where we need to update it to a different snapshot so different features aren't sharing the same version. 

By updating it to 0.1.29-SNAPSHOT it will release as 0.1.29 and we will have skipped releasing 0.1.28.

src/main/java/org/opentestsystem/ap/common/saaif/item/ItemRelease.java
```diff
@@ -3634,6 +3659,27 @@ public void setFeedback(String value) {
             @XmlAttribute(name = "filename")
             protected String filename;
 
+            @XmlAttribute(name = "xmlns:its-scripts")
+            protected String xmlNamespace;
```
bdrainer: No issues here just curious what adding this solved?  I'm interested in learning more  as I am encountering or seeing similar things as I work on EQ which includes XSDs for MathML and the EQ editor configuration.

src/main/resources/saaif_templates/mi-qti.ftl
```diff
@@ -0,0 +1,22 @@
...
+    </simpleMatchSet>
+    <simpleMatchSet>
+    <#list columns as column>
+    <#-- use character-based identifiers for columns -->
+        <simpleAssociableChoice identifier="${chars[column?index]}">
+            <p style="font-weight:bold; ">${column}</p>
+        </simpleAssociableChoice>
+    </#list>
+    </simpleMatchSet>
+</matchInteraction>
```
bdrainer: Very nice!

If you haven't seen it -https://www.imsglobal.org/question/qtiv2p1/imsqti_implv2p1.html - Brandt confirmed this is the version of QTI being used.

gradle.properties
bdrainer: I set it back to 0.1.28-SNAPSHOT and will release it now.

PR #79, opened 2017-10-15
IAT-912:  Removed the logic to detach a file.

bdrainer:


src/main/java/org/opentestsystem/ap/common/client/GitClientFactory.java
```diff
@@ -69,9 +69,17 @@ public GitClient cloneRemoteRepository(ItemBankUser userInfo, String itemId) {
...
     }
 
+    /**
+     * Attach to an item already existing on the file system.
+     *
+     * @param userInfo            The user to associate with the git requests against the repo.
+     * @param itemId              The item being attached to.
+     * @param localRepositoryPath The location of the item on the file system.
+     * @return A git client
+     */
```
bdrainer: Just added java doc to this one.  Not related to detach functionality.

PR #111, opened 2017-11-10
IAT-1020, IAT-1017 and IAT-736

alex-ponce:
* IAT-1020 (Rich Text) Space characters after rich text are stripped during preview
* IAT-1017 (HTQS) Text marked as interactive loses its styling during preview
* IAT-736 - List elements render outside of streamline content area


src/main/java/org/opentestsystem/ap/common/mapper/HtqsMapper.java
```diff
@@ -129,17 +131,21 @@ public static String mapRichTextSelectableAreas(final String modelContent, final
 
                 if (isRichTextUnderlined) {
                     final Element underlinedSpan = new Element("span");
-                    underlinedSpan.appendText(placeholder.text());
+                    underlinedSpan.appendText(placeholder.html());
```
bdrainer: nice

PR #116, opened 2017-11-16
Feature/IAT-823

hbgwhite:
changes to accommodate cc-audio linking


src/main/java/org/opentestsystem/ap/common/mapper/SaaifMapper.java
```diff
@@ -316,7 +317,19 @@ public static String mapRichTextAudio(final Item item, final String modelContent
                                 final Element source = new Element("source");
                                 source.attr("src", trimToEmpty(attachment.getFileName()));
                                 source.attr("type", MapperUtil.getAudioTypeByFile(attachment.getFileName()));
+
                                 audio.appendChild(source);
+
+                                final ItemCcAudioAssociation ccAssociation = ModelMapper.findLinkedCcFiles(item, trimToEmpty(attachment.getFileName()));
+
+                                if(ccAssociation != null) {
+                                    final Element cc = new Element("track");
```
bdrainer: I wonder if ObjectFactory has a method for creating instance of track and source.  Not a big deal.  Keep it in mind next time as that is how we have been instantiated Saaif entities.

PR #128, opened 2017-11-26
IAT-1084

hbgwhite:
When the item is not public, there is no master branch and the code removes the item repository completely. The cleanup step attempts to remove a tag on a repository that does not exist and produces a caught exception in the logs. This check eliminates the de-tag step and avoids the exception.

https://jira.fairwaytech.com/browse/IAT-1084


src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -613,7 +613,11 @@ public void rollbackEdits(final ItemBankUser user, final String itemId) {
                 gitlabClient.deleteProject(request.getProject());
             }
         } finally {
-            cleanupRequest(request);
+            if(request != null && request.isItemPublic()) {
+                cleanupRequest(request);
+            } else {
+                log.debug("skipping remove lock step for item {}, section {} (repository no longer exists)", itemId, request.getUserSection());
+            }
```
bdrainer: Looks good.  The check is only for rollback which makes sense.  I would consider a cleanupRollbackRequest method which is called here and does exactly what you are doing here.

PR #129, opened 2017-11-27
IAT-1152

hbgwhite:
Added alt text field to image resources

https://jira.fairwaytech.com/browse/IAT-1152


src/main/java/org/opentestsystem/ap/common/mapper/SaaifMapper.java
```diff
@@ -271,6 +271,7 @@ public static String mapRichTextImages(final Item item, final String modelConten
 
                         image.attr("id", lowerCase(getBaseName(trimToEmpty(imageFile.getFileName()))));
                         image.attr("src", trimToEmpty(imageFile.getFileName()));
+                        image.attr("alt", resource.getAltText() == null ? EMPTY : trimToEmpty(resource.getAltText()));
```
bdrainer: You don't need the null check because trimToEmpty does exactly that.  trimToEmpty returns EMPTY if altText is null.  It does trim off any white spaces as well.

src/main/java/org/opentestsystem/ap/common/model/ItemImageResource.java
```diff
@@ -1,12 +1,14 @@
...
 
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Data;
 import org.apache.commons.lang3.StringUtils;
 
 import static org.apache.commons.lang3.StringUtils.EMPTY;
 
 @Data
+@JsonIgnoreProperties(ignoreUnknown = true)
```
bdrainer: This is good to have.  It means removing properties won't break anything.  We should probably do this on all model objects so we are covered for property removal.

In this case though, we are adding a new property so ignoring it doesn't mean anything.  The existing items won't have the property.  When they are read altText defaults to EMPTY.  Then the next time they are written they will have the property in the item.json.

PR #132, opened 2017-12-01
IAT-942

hbgwhite:
https://jira.fairwaytech.com/browse/IAT-942

Image files require additional accessibility elements. These changes provide those elements and they are duplicated for each language. This code will have to be refactored if we ever allow them language-specific braille/audio attachments to accessibility elements for images or if we segregate our image resources by language.

Right now, all images with a production file are written. We can only write those embedded for each language if we choose, but we'd have to poll all rich text fields for each item type in that case.


src/main/java/org/opentestsystem/ap/common/mapper/SaaifMapper.java
```diff
@@ -278,6 +285,40 @@ public static String mapRichTextImages(final Item item, final String modelConten
         return doc.body().html();
     }
 
+    public static ItemRelease.Item.Content.ApipAccessibility generateSaaifAccessibility(final Item item) {
```
bdrainer: I see this is repeated for Passage and I'll guess they are identical.  There is a way to make it so there is just one method that can handle both Item and Passage.    Aside from the SAAIF Item and Passage being comprised of nested classes which was a huge mistake (that's on me), I added interfaces to the Item and Passage hierarchy.   You pass in a "supplier" that is responsible for creating Passage or Item classes like ApipAccessibility or AccessElement.  

Not a big deal, but if we get a chance it would be best to have 1 method and not 2.

src/main/java/org/opentestsystem/ap/common/model/EbsrItem.java
```diff
@@ -142,6 +142,8 @@ protected SaaifContent toSaaifContent(final String language,
             qti.setValue(itemBody);
         }
 
+        saaifContent.setApipAccessibility(SaaifMapper.generateSaaifAccessibility(this));
```
bdrainer: I see this repeated a lot as well.  What this shows to me is we need toSaaifContent on AbstractAssessmentItem.  This would be where all general assessment item content related things are done.  The best example is what you have done which is setting ApipAccessibility.  What I'm suggesting is identical to what toSaaif is doing.  All concrete types like SA, WER, MC, etc.  call "super.toSaaif(...)" as the very first thing they do in their "toSaaif(...)" method.

I don't see this being a big deal.  I think we can refactor this as time permits.  If you have time to do it now that would be cool.  If not we should do it the next time we introduce a common behavior like this.  Then it really makes sense to centralized all common/general behavior.

PR #133, opened 2017-12-01
Feature/iat 1125

alex-ponce:
* Integrated batik and jeuclid libraries to convert MathMl to SVG
* Created MathMlMapper.java
* Updated SaaifMapper.mapMathMlEquations
* Updated multiple classes to pass the ItemContext object to all mapRichTextContent calls


src/main/java/org/opentestsystem/ap/common/mapper/EbsrMapper.java
```diff
@@ -43,33 +44,37 @@
 
     public static String generateSaaifContentSingleLine(final EbsrItem item,
                                                         final EbsrItemContent content,
-                                                        final StringAssembler assembler) {
-        final String formatted = generateSaaifContent(item, content, assembler);
+                                                        final StringAssembler assembler,
+                                                        final ItemContext itemContext) {
```
bdrainer: This is fine and no need to change.  If an opportunity comes up I'll see if it makes sense to make this change...
- Remove StringAssembler from the signature as ItemContext has a StringAssembler reference.
- Add a method to ItemContext to get the StringAssembler.  ItemContext technically has AppAssembler which has StringAssembler.

src/main/java/org/opentestsystem/ap/common/mapper/MiMapper.java
```diff
@@ -63,49 +65,50 @@ public static String generateSaaifPromptSingleLine(final MiItem item, final MiIt
...
-                .collect(toList());
-
-        row.setCells(cells);
-    }
-
-    private static Table.Cell mapRichTextOnTableRowCell(final MiItem item, final Table.Cell cell) {
-        final String type = mapRichTextContent(item, cell.getType());
-        return new Table.Cell<>(type, cell.getValue());
-    }
+//    private static Table mapRichTextOnTable(final MiItem item, final Table table) {
```
bdrainer: I would delete the commented out code.  

Was this code ever used?  I'm not seeing where it is no longer being called.

PR #139, opened 2017-12-07
IAT-1218: Refactored how and where QRX files are generated.

bdrainer:
The way QRX was being generated was causing issues when committing to master.  Parallel edits would fail every single time with merge conflicts prior to this change.  That should no longer be the case.

We are generating QRX for MI, HTQO, and HTQS.  **Only HTQO** was having issues or at least that is what I noticed, so only HTQO needed fixed.


rmitchell3:
Changes look great!

PR #140, opened 2017-12-08
Feature/iat 1214

alex-ponce:
* 1214 Completed unique equation file name feature
* 1215 Reduced size of preview equation images


src/main/java/org/opentestsystem/ap/common/model/ItemContext.java
```diff
@@ -87,6 +90,9 @@ public String getQrxFileName(final String itemVersion) {
         return RepositoryUtil.getQrxFileName(itemId, itemVersion);
     }
 
+    public int getNextImageCounter() {
+        return this.imageCount.getAndIncrement();
+    }
```
bdrainer: Nice to pass it in as part of ItemContext which is already an argument on the mapper methods.

PR #147, opened 2017-12-21
Feature/iat 1256

alex-ponce:
* Fixed issue causing equations with spaces from not rendering during preview


src/main/java/org/opentestsystem/ap/common/mapper/SaaifMapper.java
```diff
@@ -1142,10 +1142,12 @@ private static String mapMathMlEquations(final String modelContent,
                     + nextFileId
                     + ".svg";
                 try {
-                    imageProps = MathMlMapper.saveMathMlToSVGFile(math, itemContext.getLocalRepositoryPath() + "/" +
-                        fileName);
+                    imageProps = MathMlMapper.saveMathMlToSVGFile(math,
+                            itemContext.getLocalRepositoryPath() + "/" + fileName);
                 } catch (Exception e) {
-                    System.out.println("Unable to create SVG reference for item: " + itemContext.getItemId());
+                    System.out.println("Unable to create SVG reference for item: " + itemContext.getItemId() +
```
bdrainer: Use the 'log' property provided by @Slf4j instead of System.out

PR #152, opened 2017-12-26
Feature/iat 1257

bdrainer:
Primary  change is using the language code in the name of the images generated from MathML.

I made the mappers instance based instead of being static methods.  No functional changes.  I did this as a small step toward a bigger refactoring of the mappers.  

We have discussed this before where an ideal direction is pulling the mappers out into a separate service.  I am not sure when that will happen but there needs to be refactoring in general for the mappers as they are getting a little cluttered.  I think the ticket to stop updating the existing SAAIF and instead regenerate it completely each time will be a good opportunity to refactor more of the mappers.


src/main/java/org/opentestsystem/ap/common/mapper/EbsrMapper.java
```diff
@@ -1,101 +1,115 @@
...
-    private static final String EBSR_CONTENT = "ebsr-content.ftl";
+    private final String EBSR_CONTENT = "ebsr-content.ftl";
 
-    private static final String PART_B_TYPE = "partBType";
+    private final String PART_B_TYPE = "partBType";
 
+    public EbsrMapper() {
+    }
+
+    public EbsrMapper(String language) {
```
rmitchell3: Recommend naming this languageCode (and everywhere else language is used) so it's clear this is the abbreviated code and not the full language name).

rmitchell3:
Nice work.  A lot of changes in this one.

src/main/java/org/opentestsystem/ap/common/mapper/EbsrMapper.java
bdrainer: Good suggestion.  I'll make this change.

bdrainer:
@rmitchell3  There were a lot of changes most of which related to making the mappers non-static.  It does trigger a little fear in me when I see this many changes however, there is some needed refactoring in the mapper code.  Mappers are a broad term for the code that transforms our model into SAAIF.

PR #157, opened 2018-01-15
Feature/iat 1293

alex-ponce:
Added ItemRepository.cloneNewItemForStimulusPreview
* Clones provided Stimulus item and it's associated files to the IVS preview directory
* Creates a new SA Item object, associates the provided Stimulus Id with this new item and saves the Saaif file to the IVS preview directory
* This method does not persist the newly created SA item to GitLab


src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -286,6 +289,39 @@ public PreviewData cloneItemForPreview(final ItemBankUser user,
...
+        log.debug("cloning new item to preview stimulus id {} for preview", stimId);
+
+        final String itemId = saaifIdGenerator.generateTransactionId();
+        final SaItem item = (SaItem) itemFactory.newItem(itemId, ItemConstants.ItemType.TYPE_SA);
+
+        final ItemContext itemContext = new ItemContext(itemId, this.appAssembler, repositoryPath);
+        final ItemRelease itemRelease = item.newSaaifItem(bankKey, itemContext);
+
+        item.getCore().setStimulusId(stimId);
+        item.getCore().getEn().setPrompt("System generated Short Answer used to preview Stimulus");
```
bdrainer: Nice.  I have been thinking about setting the prompt since our conversation Friday.  This is the easiest way.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -286,6 +289,39 @@ public PreviewData cloneItemForPreview(final ItemBankUser user,
...
+
+        final ItemRelease updatedSaaifItem = item.updateItemForPreview(
+                item.toSaaif(itemRelease, itemContext),
```
bdrainer: I would have to look closer but I wonder if calling item.toSaaif() is necessary.  Can you just pass itemRelease to item.updateItemPreview?

alex-ponce: item.toSaaif was needed to get the attributes list update for the stimulus Id. I noticed that calling setAssociatedPassage only updates the <associatedPassage> element but not the attributes. BTW this helped confirm IVS only looks at the attributes to find an association with a Stimulus.

rmitchell3:
This all looks good.  Nice work!

PR #162, opened 2018-01-30
Feature/iat 1340

alex-ponce:
* Removed Grade8Or11OnlyAlgebraDescriptor metadata field
* Added AlgebraDescriptor1 metadata field
* Added AlgebraDescriptor2 metadata field
* Added MetaDataFixTest.java to consolidate all metadata field checks that need to happen during UAT deployment of the Sprint 2 deliverable


src/main/java/org/opentestsystem/ap/common/model/ItemMetadata.java
```diff
@@ -36,7 +36,9 @@
 
     private String grade = EMPTY;
 
-    private String grade8Or11OnlyAlgebraDescriptor = EMPTY;
+    private String AlgebraDescriptor1 = EMPTY;
```
bdrainer: You must be a .NET developer which explains the capital on a property name.  Java the first letter is lower case.

src/test/java/org/opentestsystem/ap/common/it/MetaDataFixTest.java
```diff
@@ -0,0 +1,199 @@
...
+        if (isNotEmpty(items)) {
+            System.out.println("3. Read items and check metadata\n");
+            items.stream().forEach(item -> {
+                for (String branch : item.getBranches()) {
+                    final String migrations = requiredMigrations(item, branch);
+                    if (!migrations.equals("")) {
+                        item.setBranch(branch);
+                        item.setNeedsMigration(true);
+                        item.setMigrationsNeeded(migrations);
+                        // break;
```
bdrainer: If you don't 'break' out of the loop when the item + branch requires migration then you will check the same item multiple times on all its branches.  If the master branch needs migrated and the create branch needs migrated you will check them both, but you will only record the last one checked.  The reason I was breaking out of the for  loop is to avoid checking every branch for the item.

PR #172, opened 2018-02-19
Feature/iat 1424

bdrainer:
Hi @rmitchell3 and @alex-ponce,

I added property **version** that is persisted with the item.  It is a sibling property of Item.id and Item.type.

I also added a property called **modelVersion** to Item.  It is not persisted in the item.json.  It is used when saving an item.  I made this value a constant.  I wasn't sure why it needs to be a configuration property.  Let me know if you have thoughts on this.  The Wiki calls out making this a configurable property.  I could remove the static property and use a config property.  Easy to do but there is something about using a config property I don't like.


src/main/java/org/opentestsystem/ap/common/client/GitClient.java
```diff
@@ -282,6 +282,9 @@ void mergeSectionBranch(final String section) {
      * @param commitMessage The commit message associated with this action.
      */
     public void updateItem(final Item item, final String section, final String commitMessage) {
+        // ensure the item version is up to date
+        item.setVersion(item.getModelVersion());
```
bdrainer: Here the item is being saved.  The static 'modelVersion' property is used to set the item's 'version' property.  This is the only place the 'version' is updated.  It is not updated when read.  

You can see how easy it would be to change this to use a configuration property instead of the 'modelVersion' property on Item.

src/main/java/org/opentestsystem/ap/common/model/AbstractItem.java
```diff
@@ -54,6 +54,8 @@
 
     private TextToSpeech textToSpeech = new TextToSpeech();
 
+    private String version = EMPTY;
```
bdrainer: Item.version is set to "" when first instantiated.  When reading the item if it does not have a 'version' set the default is "".  This how we can tell if an item has never had its 'version' property set.  If the version property is equal to empty string then it has never been set.

src/main/java/org/opentestsystem/ap/common/model/Item.java
```diff
@@ -9,31 +9,32 @@
...
+    @JsonSubTypes.Type(value = EbsrItem.class, name = "ebsr"),
+    @JsonSubTypes.Type(value = TiItem.class, name = "ti"),
+    @JsonSubTypes.Type(value = EqItem.class, name = "eq"),
+    @JsonSubTypes.Type(value = HtqoItem.class, name = "htqo"),
+    @JsonSubTypes.Type(value = HtqsItem.class, name = "htqs"),
+    @JsonSubTypes.Type(value = GiItem.class, name = "gi")
 })
 public interface Item<T extends ItemCore, S extends ItemContent> {
 
+    String MODEL_VERSION = "1";
```
bdrainer: Here is the hard coded model version.  I am starting it at 1 and assuming anything prior to this is version 0 (or whatever comes before version 1.  Maybe this should be version 2 and everything before this is version 1.  Let me know if you have thoughts on this.

PR #179, opened 2018-02-27
Feature/iat 1330

alex-ponce:
ASL Attachments require a <source> element in the resource Saaif XML. To accomplish this the following changes were made:
* Added SaaifAttachmentSource.java
* Integrated SaaifAttachmentSource to SaaifItemSupplier.java and SaaifPassageSupplier.java
* Integrated SaaifAttachmentSource to ItemRelease.java


src/main/java/org/opentestsystem/ap/common/mapper/SaaifMapper.java
```diff
@@ -1225,20 +1226,38 @@ public SaaifAttachment newSaaifAttachment(final Attachment attachment,
...
+                                              final String subType,
+                                              final String section) {
         final SaaifAttachment saaifAttachment = supplier.newAttachment().get();
 
         saaifAttachment.setType(type);
         saaifAttachment.setSubtype(subType);
         saaifAttachment.setFile(fileName);
 
+        if (section != null) {
+            if (section.equalsIgnoreCase("ASL")) {
```
bdrainer: You could turn this into one 'if' with ```if ("ASL".equalsIgnoreCase(section)) ```

alex-ponce: I can (should), I had logic that also interrogated "Braille" and didn't clean up. I will change it before you merge

src/main/java/org/opentestsystem/ap/common/model/AbstractItem.java
```diff
@@ -328,10 +328,10 @@ protected void updateSaaifTranslations(final ItemRelease itemRelease, final Item
 
         final List<SaaifAttachment> saaifBraille = saaifMapper().newSaaifAttachment(braille, SECTION_BRAILLE, getSupplier());
 
-        final Attachment emptyBraille = Attachment.emptyBraille();
-
-        saaifBraille.add(saaifMapper().newSaaifAttachment(
-                getSupplier(), emptyBraille.getFileName(), emptyBraille.getUploadedDate(), null, null));
+//        final Attachment emptyBraille = Attachment.emptyBraille();
+//
+//        saaifBraille.add(saaifMapper().newSaaifAttachment(
```
bdrainer: The emptyAsl and emptyBraille were used to avoid merge conflicts.  They were a divider.  It divided ASL from Braille.   

Technically only one of the dividers divides asl from braille, I think the braille one.  The idea being all asl sorts above braille, then comes the braille divider, and after it are listed all the braille attachments.  

I wonder if having an empty divider breaks anything down stream.  What I do know is without it there is a chance of a merge conflict.  One user can add ASL at the same time a different user is adding Braille.  If there is no divider there will be a merge conflict.

alex-ponce: OK, will put back the empty attachments

PR #187, opened 2018-03-20
Feature/iat 1488

alex-ponce:
Added updatedDate attribute to the following item types:
* SaItem
* StimItem
* WerItem
* TutItem
* MsItem
* McItem
* HtqsItem
* Htqoitem
* MiItem
* TiItem
* GiItem
* EqItem
* EbsrItem

* Each item type, e.g SaItemContent class overrides the "equals" method
* Each item type, e.g. SaItem class overrides the "setCoreUpdatedDate" and "setTranslationUpdatedDate" methods


src/main/java/org/opentestsystem/ap/common/model/eqconfig/EqEquationEditorVariableRowConfiguration.java
```diff
@@ -5,8 +5,11 @@
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import lombok.Data;
+import lombok.EqualsAndHashCode;
+import org.opentestsystem.ap.common.exception.SystemException;
 
 @Data
+@EqualsAndHashCode
```
bdrainer: Good use of annotation.  The whole less is more approach.

src/main/java/org/opentestsystem/ap/common/model/EbsrItem.java
```diff
@@ -78,6 +80,26 @@ public void setCore(EbsrItemCore core) {
...
+        EbsrItemCore ebsrItemCore = (EbsrItemCore) core;
+        Date updatedDate = new Date();
+        if( this.core.getEn().equals(ebsrItemCore.getEn())) {
+            updatedDate = ebsrItemCore.getEn().getUpdatedDate();
+        }
+        this.core.getEn().setUpdatedDate(updatedDate);
+    }
+
+    @Override
+    protected <T extends ItemCore, S extends ItemContent> void setTranslationUpdatedDate(S content) {
```
bdrainer: I wonder if implementing this method on all types was necessary.  The arg passed in is of type ItemContent.  You then cast it to the specific type which is what I think is unnecessary.   I'm thinking in AbstractItem you implement  "setTranslationUpdatedDate".  You would need to make sure "setUpdatedDate" is exposed on ItemContent.

"setCoreUpdatedDate" was harder because ItemCore does not expose ItemContent. If it did then we would only have needed one implementation of it.

PR #197, opened 2018-04-23
Feature/iat 1484

alex-ponce:
Multiple metadata field name changes defined here:

https://confluence.fairwaytech.com/display/IAT/2.2.4+-+Sync+IAT+with+canonical+fields+from+IMRT+data+dictionary


src/main/java/org/opentestsystem/ap/common/model/ItemMetadata.java
```diff
@@ -87,9 +79,9 @@
 
     private String primaryMathematicalPractice = EMPTY;
 
-    private String primaryTargetSet = EMPTY;
+    private String primaryTarget = EMPTY;
```
millerts: This will break IMRT.  Please let us know when this is released so IMRT can make the appropriate changes to consume this new value.

We will also need to know when this is in development.

millerts: I spoke with @rmitchell3 about this offline.  IMRT uses the model objects within AP_Common to deserialize the item.json file as it ingests it.  This means that IMRT currently uses `primaryAssessmentTarget`.  We do this so we can derive certain values the client wants to be searchable.  If the model (item.json) changes and IMRT does not have the new Java model it won't deserialize the object properly.

Long story short once this is in development IMRT can be updated to ingest the new target field appropriately.

PR #208, opened 2018-05-23
Feature/iat 1836

bdrainer:
A lot of changes below are from organizing imports on the model package

See Attachment.java - this is where I'm hard coding the date to the ones currently being used on production.  I chose to stick with those so there won't be any changes during migration unless that value is off somehow.  

See JsonModelAssembler - this is where the ObjectMapper has been configured to sort properties alphabetically.  When the Item is written to json the properties will be ordered this way.  The exception is item ID, item type, and migration version are the first listed for item.


src/main/java/org/opentestsystem/ap/common/model/Attachment.java
```diff
@@ -1,22 +1,21 @@
...
  * An attachment associated with an item.
  */
 @Data
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Attachment {
 
-    private static final Date EMPTY_ATTACHMENT_ASL_DATE = new GregorianCalendar(1975, 3, 21).getTime();
-    private static final Date EMPTY_ATTACHMENT_BRAILLE_DATE = new GregorianCalendar(1976, 3, 21).getTime();
+    private static final Date EMPTY_ATTACHMENT_ASL_DATE = new Date(167270400000L);
+    private static final Date EMPTY_ATTACHMENT_BRAILLE_DATE = new Date(198892800000L);
```
bdrainer: Hard cod the date here.  These are the values currently set on production.  Using the same value means nothing will change on production, unless it is different than this value.

src/main/java/org/opentestsystem/ap/common/saaif/JsonModelAssembler.java
```diff
@@ -19,6 +22,11 @@
 
     final ObjectMapper mapper = new ObjectMapper();
 
+    public JsonModelAssembler() {
+        mapper.configure(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY, true);
+        mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
```
bdrainer: Here we configure the properties to be order alphabetic.  And also not to fail on unknown properties which is a good addition as previously we were annotating the model classes, but have probably missed some places, so this configuration covers us.

PR #212, opened 2018-06-04
Feature/imrt 360

millerts:


src/main/java/org/opentestsystem/ap/common/model/FieldTestData.java
```diff
@@ -0,0 +1,162 @@
...
+    private String countSelectF = EMPTY;
+    private String countSelectG = EMPTY;
+    private String countSelectH = EMPTY;
+    private String countTitle1 = EMPTY;
+    private String countWhite = EMPTY;
+    private String difCatAsianvsWhite = EMPTY;
+    private String difCatBlackvsWhite = EMPTY;
+    private String difCatFvsM = EMPTY;
+    private String difCatHispanicvsWhite = EMPTY;
+    private String difCatIEPvsNonIEP = EMPTY;
```
millerts: Let me know if this should be `difCatIepVsNonIep`.  With all the acronyms within the field test data I wasn't sure the best naming.

src/test/java/org/opentestsystem/ap/common/it/IntegrationTest.java
```diff
@@ -237,6 +243,46 @@ public void createItem() throws IOException {
         new ObjectMapper().writerWithDefaultPrettyPrinter().writeValue(out, itemResponse);
     }
 
+    @Ignore
+    @Test
+    public void createItemWithTestAdministrationData() {
```
millerts: This is the only test around this data.  If there is more tests I should add please let me know.

src/main/java/org/opentestsystem/ap/common/model/FieldTestData.java
rmitchell3: I'd recommend acronyms only capitalize first letter.  "vs" should have capital 'v'.  So this one should be:  difCatIepVsNonIep.  This is so cryptic I have no idea what it means.

I also think we should expand all acronyms and use no abbreviations (maybe we could use "Vs" as short for "Versus" but I'd be OK spelling that out too.

src/main/java/org/opentestsystem/ap/common/model/FieldTestData.java
```diff
@@ -0,0 +1,162 @@
...
+    private String exactRaterAgreement = EMPTY;
+    private String excludedFromAllAnalysis = EMPTY;
+    private String excludedFromIRTAnalysis = EMPTY;
+    private String irtA = EMPTY;
+    private String irtB = EMPTY;
+    private String irtC = EMPTY;
+    private String irtFit1 = EMPTY;
+    private String irtFit2 = EMPTY;
+    private String irtStep1 = EMPTY;
+    private String rtStep2 = EMPTY;
```
rmitchell3: should be irtStep2

millerts:
All @rmitchell3 comments have been addressed.  We paired on the names and abbreviations are not present.  There are two TODO's present because they're reliant on the client answer the last two questions for abbreviations

PR #215, opened 2018-06-13
IAT-1650: Refactored the item history lookup.  The lookup now pages t…

bdrainer:
…hrough until all commits are retrieved.

This one is interesting.  I had to add a method to GitlabAPI which uses Feign to do all the heavy  lifting.  We use the GitlabApi class when the Gitlab client we use does not support what we need, and in this case the Gitlab client does not support paging through the commits.

While working on this I noticed a few unused pieces of code so I removed them.  For example, we used to filter the commit messages removing any that came from the validation service.  The validation service no longer commits to the item history, so there is no need to filter anymore.


src/main/java/org/opentestsystem/ap/common/client/GitlabApi.java
```diff
@@ -62,9 +63,11 @@ ItemDiff getItemDiffWithRange(@Param("projectId") final String projectId,
...
                                @Param("perPage") final Integer perPage);
 
-    @RequestLine("GET /projects/{projectId}/repository/commits?ref_name={branchName}")
-    List<Commit> getBranchCommits(@Param("projectId") final Integer projectId,
-                                  @Param("branchName") final String branchName);
+    @RequestLine("GET /projects/{projectId}/repository/commits?ref_name={branchName}&page={page}&per_page={perPage}")
+    List<ItemHistory> getBranchCommits(@Param("projectId") final Integer projectId,
+                                       @Param("branchName") final String branchName,
+                                       @Param("page") final Integer page,
+                                       @Param("perPage") final Integer perPage);
```
bdrainer: This is the method allowing us to pass in the page number and page size.

src/main/java/org/opentestsystem/ap/common/config/ItemBankProperties.java
```diff
@@ -77,4 +74,6 @@
     private String reportRepoName = "ItemBankReports";
 
     private String systemVersion;
+
+    private int commitsPageSize = 100;
```
bdrainer: Picked 100 as the default page size.  The amount of data returned isn't a lot (IMO) so 100 seems ok.  If it needs to change we can override with a config setting.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -748,6 +714,36 @@ public String findItemHistoryCsvString(final String itemId) {
...
     // ------------------------------------------------------------------------
 
+    /**
+     * Finds all commits for the item off of its master branch, so only commits from the master branch are returned.
+     *
+     * @param itemId The unique item identifier.
+     * @return All item commits from its master branch.
+     */
+    @VisibleForTesting
+    List<ItemHistory> findItemCommits(final String itemId) {
```
bdrainer: This is the method does all the work to get the full list of commits.

PR #242, opened 2018-07-27
IAT-2116 : ASL and Braille Saaif attachment refactoring

alex-ponce:
During testing it was discovered that ASL Saaif attachments were not being processed correctly. Prior to this change there was a one-to-one relationship between an Asl attachment in item.json and an Saaif attachment in the items xml file. This is correct for Braille attachments however not for Asl.
All production items contain just one "ASL" attachment and one or more <source> elements. For Asl the <source> elements have a one-to-one relationship with Asl attachments in item.json.

Below is an example of the expected results when an item has 2 braille and 2 asl attachments

`<attachmentlist>`
`        <attachment id="braillefile1" type="BRF" subtype="contracted" file="passage_1007_enu_contracted.brf" />`
`        <attachment id="braillefile2" type="BRF" subtype="uncontracted" file="passage_1007_enu_uncontracted.brf" />`
`        <attachment id="aslfile1" type="ASL" subtype="STEM" file="passage_1007_ASL_STEM.mp4">`
`          <source src="passage_1007_ASL_STEM.webm" type="video/webm" />`
`          <source src="passage_1007_ASL_STEM.mp4" type="video/mp4" />`
 `       </attachment>`
`      </attachmentlist>`


bdrainer:
Good fix and good catch.

src/main/java/org/opentestsystem/ap/common/saaif/item/ItemRelease.java
```diff
@@ -2740,7 +2742,7 @@ public void setTextToSpeechPronunciationAlternate(String value) {
                     protected String pass;
 
                     @XmlElement()
-                    protected ItemRelease.Item.Content.Attachmentlist.Attachment.AttachmentSource source;
+                    protected List<ItemRelease.Item.Content.Attachmentlist.Attachment.AttachmentSource> sources;
```
bdrainer: Required changes to SAAIF.  I bet it was List initially and I thought it didn't need to be so I changed it to single instance.

src/main/resources/SAAIF-Metadata.xsd
```diff
@@ -48,7 +48,7 @@
                 <xs:element type="xs:string" name="EducationalDifficulty"/>
                 <xs:element type="xs:string" name="DepthOfKnowledge"/>
                 <xs:element type="xs:string" name="EligibleAssessmentType" maxOccurs="unbounded" minOccurs="0"/>
-                <xs:element type="xs:string" name="SpecificationVersion"/>
+                <xs:element type="xs:string" name="SpecificationsVersion"/>
```
bdrainer: nice

PR #250, opened 2018-08-22
Feature/iat 2096

millerts:


src/main/java/org/opentestsystem/ap/common/model/AbstractItemCore.java
```diff
@@ -21,11 +22,23 @@ protected void setProperties(AbstractItemCore copy) {
...
         }
     }
+
+    @Override
+    public List<ContentUpdateNeed> getContentUpdateNeeds() {
+        return contentUpdateNeeds;
+    }
+
+    @Override
+    public void setContentUpdateNeeds(List<ContentUpdateNeed> contentUpdateNeeds) {
```
millerts: @rmitchell3 , reading the content update need confluence requirements pages it seems like they can show up on any item.  If they are only valid for certain item types please let me know and I'll update this PR.

millerts: Had an offline discussion with Ray.  This is for all items so placement is correct within the object hierarchy.

PR #252, opened 2018-08-27
Add additional glossary information

millerts:


src/main/java/org/opentestsystem/ap/common/model/Glossary.java
```diff
@@ -10,9 +12,23 @@
 @Data
 @JsonIgnoreProperties(ignoreUnknown = true)
 public class Glossary {
-
     private List<GlossaryTerm> terms = new ArrayList<>(0);
 
+    @JsonProperty("isEnglishGlossaryProvided")
+    private String englishGlossaryProvided = "No";
```
millerts: The field values here are set to "No" as mentioned in https://confluence.fairwaytech.com/display/IAT/Add+fields+to+Glossary+tab.  This is the only place that I saw where we set the default values in Java code which is why I asked if it should be defaulted here or somewhere else in IAT.

PR #257, opened 2018-08-31
Feature/iat 2082

millerts:
Removes .wav and adds m4a as a valid glossary extension


src/main/java/org/opentestsystem/ap/common/itembank/attachments/AttachmentValidator.java
```diff
@@ -110,6 +111,18 @@ public void validateAttachmentRequest(final String itemId, final AttachmentReque
...
 
+    /**
+     * Depending on the section the file name is validated based on a known pattern. Audio files have one pattern and
+     * braille files have a different pattern for example.
+     *
+     * @param fileName The file name to validate
+     * @param section  The section the file is associated with
+     * @throws ValidationException thrown if the file name is invalid
+     */
+    public void validateFileName(final String fileName, final String section) throws ValidationException {
```
millerts: I added this convenience method since validateFileName's third argument is option as in it should be blank.

PR #301, opened 2018-10-08
Process wordlist during rich text mapping

alex-ponce:
* Added a WordlistreleaseType param to all SaaifMapper.mapRichTextContent method calls
* During glossary term mapping, the item.json glossary term is used as a lookup in the recently added WordlistreleaseType
* This lookup prevents issues caused when a glossary term is deleted in item.json


alex-ponce:
Confirmed with @bdrainer that approach is correct

PR #304, opened 2018-10-10
Added a check for the create file and add it on commits for newly cre…

millerts:
…ated items.  This is for items that were created prior to the "create" file addition


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -150,6 +150,29 @@ public boolean doesAttachmentExist(String itemId, String branch, String fileName
...
+     * @param newFileName      the new file name
+     * @return {@code true} if the file was moved properly
+     */
+    public boolean renameAttachment(String itemId, String branch, String originalFileName, String newFileName) {
+        if (!doesAttachmentExist(itemId, branch, originalFileName)) {
+            return false;
+        }
+
+        String originalFileKey = dataStoreUtility.replaceFileKey(itemId, branch, originalFileName);
+        String newFileKey = dataStoreUtility.replaceFileKey(itemId, branch, newFileName);
```
bdrainer: why did I call the util method 'replaceFileKey', I bet I meant to rename that to 'generateFileKey', we can rename it later

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -150,6 +150,29 @@ public boolean doesAttachmentExist(String itemId, String branch, String fileName
...
+
+        amazonClient.copyFile(originalFileKey, newFileKey);
+        amazonClient.deleteFile(originalFileKey);
```
bdrainer: I wondered if we should delete. We should   The sooner we can get things out of S3 the better in general.  GitLab will retain the deleted attachment anyway since we never delete from there.  Just wanted to mention that.

src/main/java/org/opentestsystem/ap/common/datastore/client/AmazonClient.java
```diff
@@ -103,6 +103,17 @@ public InputStream downloadFile(String fileKey) {
...
 
+    /**
+     * Delete an S3 object by key
+     *
+     * @param fileSourceKey the unique key for the file
+     */
+    public void deleteFile(String fileSourceKey) {
+        log.info("delete file {}", fileSourceKey);
+
+        client.deleteObject(awsProps.getBucketName(), fileSourceKey);
```
bdrainer: nice

PR #333, opened 2018-12-11
EID-25: enhance metadata create to include enhanced IDs

gregcharles:


src/main/java/org/opentestsystem/ap/common/saaif/transformer/SaaifMapper.java
```diff
@@ -293,6 +300,14 @@ public SmarterAppMetadata mapToMetadata(Item item, List<String> enemyItems) {
         return metadata;
     }
 
+    private boolean hasEnhancedIds(ServerManaged modelServerManaged) {
+        return modelServerManaged != null &&
```
millerts: modelServerManaged is never null.  If null it seems like this is Exceptional.

PR #362, opened 2019-02-07
iat 2708 - Validation: Score points error regarding quotes is generated against items with points assigned

alex-ponce:
Saaif Xml files require `"` characters to be present in element values. e.g.
`<ScorePoints>"0, 1"</ScorePoints>`
By default, the XML Marshaller used in IAT converts `"` characters to `&quot;` e.g.
`<ScorePoints>&quot;0, 1&quot;</ScorePoints>`

To avoid this, an `SaaifCharacterEscapeHandler` class which implements `CharacterEscapeHandler` was introduced. When set as a property of the Marshaller it produces the expected result.

Note: To isolate this change it was only added to the `SaaifMetadataAssembler` which will only affect the contents of metadata.xml


src/main/java/org/opentestsystem/ap/common/saaif/transformer/SaaifMapper.java
```diff
@@ -2385,6 +2386,14 @@ private Document newDocument(String modelContent) {
         return doc;
     }
 
+    private String getScorePointsMetadataValue(String scorePoints) {
+        String value = "";
+        if (scorePoints != null) {
+            value = String.format("\"%s\"", scorePoints);
```
bdrainer: I see, we are adding the quotes and then the escape handler **is not** converting the double quotes to the escape character.  Makes sense.

src/test/java/org/opentestsystem/ap/common/saaif/transformer/SaTransformerTest.java
```diff
@@ -200,11 +200,28 @@ public void itShouldCreateTranslationTTSSaaifContent() throws Exception {
...
+        try (FileInputStream fis = new FileInputStream(SA_MATH_JSON)) {
+            SaItem item = (SaItem) testUtil.getAssembler().getJsonModelAssembler().readItem(fis);
+
+            ItemContext ctx = new ItemContext(item.getId(), item.getLegacyId(BANK_KEY), BANK_KEY, testUtil.getAssembler(), testUtil.getLocalRepoDir());
+
+            saTransformer.generateSaaifFiles(item, BANK_KEY, ctx);
+
+            String metadataXml = FileUtils.readFileToString(ctx.getLocalMetadataFilePath().toFile(), "UTF-8");
+
+            Assert.assertTrue(metadataXml.contains("<ScorePoints>\"0, 1\"</ScorePoints>"));
```
bdrainer: Who doesn't love a good unit test.  Very nice.

PR #387, opened 2019-03-14
Hotfix/0.4.127.1

millerts:


bdrainer:
Nice changes.  Get rid of the relationship management Hibernate was doing and just manage it yourself with queries.  You still get to use the Spring Data stuff so you didn't have to go full JdbcTemplate.  This should do it in my opinion and if not it still a huge step in the right direction.

PR #427, opened 2019-04-19
IAT-3002 - Refactored unzip method

alex-ponce:
* Removed unzip method from MigrationFileUtil
* Refactored unzip method in ZipUtil to be compatible with different types of archive files


src/main/java/org/opentestsystem/ap/common/util/ZipUtil.java
```diff
@@ -29,10 +30,21 @@ public static void unzip(Path zipFile, Path unzipPath) {
...
+                    // Certain archive files do not return individual directory entries
+                    // Instead they return the fully qualified directory and filename in each entry
+                    // e.g. /Item-200-42403//Item_42403_v16.qrx
+                    // Create all directories prior to calling unzipFiles
+                    File newFile = new File(unzipPath.toString() + File.separator + entry.getName());
+                    new File(newFile.getParent()).mkdirs();
+                    // entry value: item-200-42403/Item_42403_v11.qrx
+                    unzipFiles(zipInputStream, newFile.toPath());
+                }
+                else {
```
millerts: Small nit, can you make the else be on the same line as the closing bracket for the else if?

PR #464, opened 2019-05-19
Bugfix/iat 2993

millerts:


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -388,6 +388,11 @@ private void updateAssociatedTutorialId(ItemBankUser user,
                                             List<String> itemIds,
                                             String message) {
         ItemEntity entity = this.dataManager.findLastMaster(tutorialId);
+
+        if(entity == null) {
```
millerts: I feel this logic and the entire class should not be in common but in bulk service.  I put this change in common since this class is already here.

PR #481, opened 2019-06-27
iat 3124 - Write created by full name and commit message to the item table

alex-ponce:
*For bulk operations* 
* Added updateOwnerFullName to the ItemBulkUpdate model class. Use this value to populate the item table. The commit message is already part of the operation.

*For TIMS operations*
* Use the existing ItemBankUser object to retrieve the users full name. The commit message is already part of the operation


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -103,7 +103,7 @@ public void lock(String itemId) throws ItemBulkUpdateCheckedException {
     // ------------------------------------------------------------------------
 
     @Transactional
-    public void updateAsl(String updateOwner, ItemBulkUpdateAsl.AslFilePair pair) {
+    public void updateAsl(String updateOwner, String updateOwnerFullName, ItemBulkUpdateAsl.AslFilePair pair) {
```
millerts: If we decide we need even more user information when saving data I think we should consider creating a User object that can contain this data and have it passed around instead of Strings.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -95,7 +95,7 @@ public Long deleteItem(String itemId) {
     }
 
     @Transactional
-    public ItemResponse beginItemEdit(ItemBankUser user, String itemId, String branch) {
+    public ItemResponse beginItemEdit(ItemBankUser user, String itemId, String branch, String commitMessage) {
```
millerts: Do we ever expect this to be null or blank after this change?  And if not, should we throw an IllegalArgumentException assuming it is a software bug?

PR #482, opened 2019-07-03
Feature/iat 3155

bdrainer:
**This is a big change**

The main change is 
- IMS commits attachments to the versioned S3 bucket as part of committing changes from a branch to master
- The history service.  Currently the history service is a consumer of item events.  It downloads the version attachments IMS just created, generates the SAAIF, zips and stores that in S3, and finally inserts an ITEM_HISTORY record.

Other changes are
- Renamed a few classes, the main one is AmazonClient is now S3BucketManager - **this caused a lot of files to change**
- Added StopWatchLogger.  It uses the Apache StopWatch.  I wrapped that to log an info message with a consistent/searchable prefix.
- Started down the road of using the zips for preview and validation.  I put a lot of performance logging in the preview code.


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -302,20 +326,27 @@ public void patchItem(ItemBankUser user,
                           String... restrictions) {
         log.info("user {} is changing field '{}' on item item {} ", user.getUserName(), propertyPath, propertyValue);
 
-        if (Objects.isNull(this.dataManager.lockMaster(itemId))) {
-            throw new ResourceNotFoundException("Item " + itemId + " not found");
-        }
+        // always lock
+        this.lockThrowResourceNotFound(itemId);
```
millerts: I realize this is a bigger change, but I wonder if we should have a single method entry point for patching that could take an enum for the type of operation being done.  And we could have some generic object as the input.  That single entry point could the lock then delegate to the patches.  My worry is that someone will add a method to bulk and forget to add this lock.

src/main/java/org/opentestsystem/ap/common/datastore/AsyncDataStorePreviewManager.java
```diff
@@ -50,14 +51,18 @@ public AsyncDataStorePreviewManager(ItemBankProperties properties,
     }
 
     private CompletableFuture<PreviewData> generatePreviewData(Item item, String branch, String bankKey, Path itemFolder) {
-        this.attachmentManager.downloadAttachmentsForPreview(item.getId(), branch, itemFolder);
+        log.info("generate preview data: {}, branch {}, bank key {}", item.getId(), branch, bankKey);
+        StopWatchLogger watch = StopWatchLogger.start();
```
millerts: Cool.  I've never used this.  Seems simpler than my Instant logging in IMRT.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -82,7 +108,36 @@ public InputStream getAttachment(String fileKey) {
     public List<String> findAttachments(String itemId, String branch) {
         String branchKey = dataStoreUtility.generateFolderKey(itemId, branch);
         log.info("Find branch attachments at {}", branchKey);
-        return amazonClient.listFiles(branchKey);
+        return s3BucketManager.listFiles(branchKey);
+    }
+
+    /**
```
millerts: Thanks for adding JavaDocs

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
```diff
@@ -6,41 +6,57 @@
...
 /**
- * Amazon client.
+ * Client for interacting with S3.
  */
 @Slf4j
-public class AmazonClient {
+public class S3BucketManager {
 
-    private final ItemBankProperties.AwsProps awsProps;
+    protected final ItemBankProperties.AwsProps awsProps;
```
millerts: Suggest leaving this private and add setter/getter if we need to override it.

src/main/java/org/opentestsystem/ap/common/history/ItemHistoryEventHandler.java
```diff
@@ -0,0 +1,182 @@
...
+                    log.info("[HISTORY] {} - item {} | master commit {} | branch commit {} | username {}",
+                        StringUtils.upperCase(eventType),
+                        masterCommit.getItemId(),
+                        itemEvent.getMasterCommitId(),
+                        itemEvent.getBranchCommitId(),
+                        itemEvent.getUser().getUserName());
+                    // commit and patch are the same
+                    this.saveHistoryRecord(itemEvent, masterCommit);
+                } else {
+                    log.info("[HISTORY] SKIP {} - item {} is being edited | master commit {} username {}",
```
millerts: Should this be at debug since we're skipping?  Much like you have on 99.

bdrainer: yes, I'll change that

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
bdrainer: sounds good, I'll make that change

src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
bdrainer: It's a good point. There are areas we could remove redundant code and areas to help ensure things like patch or commit behave the same every where and only need changed in a single place (ims, ibms, migration, etc.).

PR #485, opened 2019-07-16
iat-3202 - Fix issue where the Migration application doesn't find the import files

alex-ponce:


src/main/java/org/opentestsystem/ap/common/saaif/mapper/util/MigrationFileUtil.java
```diff
@@ -564,16 +535,23 @@ public static File getSourceDirectoryPath(String sourceDirectory, String itemDir
...
+                    if (dirParts[2].equals(itsId)) {
+                        itemDir = rootDirectory.resolve(file.getName());
+                        break;
+                    }
                 }
             }
         }
-
-        return rootDirectory;
+        if (Objects.isNull(itemDir)) {
```
millerts: Not a critical thing but recommend using Optional for nullable items.  It isn't necessarily worth it here since this is a small method but wanted to mention it for future reference.

It'd change it to this:
```
Optional<Path> itemDir = Optional.empty();
...
itemDir = Optional.of(rootDirectory.resolve(file.getName());
...
if (!itemDir.isPresent())
...
return itemDir.get();
```

src/test/java/org/opentestsystem/ap/common/saaif/mapper/util/MigrationFileUtilTest.java
```diff
@@ -0,0 +1,77 @@
...
+            Path tempDir = Files.createTempDirectory("test");
+            Files.createDirectories(tempDir.resolve(RandomStringUtils.randomAlphabetic(RANDOM_DIR_CHAR_LENGTH)));
+            Files.createDirectories(tempDir.resolve(validDirectoryName));
+            Files.createDirectories(tempDir.resolve(RandomStringUtils.randomAlphabetic(RANDOM_DIR_CHAR_LENGTH)));
+            Files.createDirectories(tempDir.resolve(RandomStringUtils.randomAlphabetic(RANDOM_DIR_CHAR_LENGTH)));
+
+            Path itemDirectory = MigrationFileUtil.getItemSourceDir(tempDir, itsId);
+
+            Assert.assertEquals(itemDirectory.getFileName().toString(), validDirectoryName);
+        } catch (IOException e) {
```
millerts: Rather than catch and log I feel all these tests should just throw.  We'll want to see the failures, and tracking down the logs when building common will be hard to do.

alex-ponce:
Todd's suggestions were correct and easy to implement.

PR #486, opened 2019-07-17
IAT-3020: Add 'X' to certain target value for Math items - changed th…

bdrainer:
…e model mapper so it de-dupes Xs on the three 'target' properties.  

We need this change so migration don't re-introduce the multiple Xs.  We don't need a migration, so there will be no accompanying migration in the migration app to change all items where their 'target' metadata properties are multiple Xs.


src/main/java/org/opentestsystem/ap/common/saaif/mapper/model/StandardIdProps.java
```diff
@@ -71,23 +75,27 @@ public StandardIdProps loadFromValue(String idType, String value) {
         return this;
     }
 
-    private void setTargetAndGrade(String value) {
+    void setTargetAndGrade(String value) {
```
millerts: Can you add the annotation `@VisibleForTesting` to both these methods since they're only package private for testing purposes?  The annotation is documentation based and does nothing to the functional code.

bdrainer:
Thanks for the review and changes.  I realize now I could have left those methods 'private' as I tested it using the 'public' methods.  I did intend to test it directly with those methods but didn't.

PR #487, opened 2019-07-19
iat 3168 - Add an endpoint to get SAAIF data for an item

alex-ponce:
Added end-point that returns a zip file containing all Saaif and attachments for the requested item

`/api/v1/items/67861?format=saaif`


alex-ponce:
Closing PR per team decision to leave this class in the Item Translation Service

PR #488, opened 2019-07-23
Feature/iat 3181

bdrainer:
There are three main things in the changes below and a minor third change.

1. The primary change is preview and validation no longer use GitLab.  It came down to the historic preview being the challenging part.  See: DataStorePreviewManager, AsyncDataStorePreviewManager, and PreviewManager.  At standup, I'll walk us through these classes to point things out.

2. Removing the dependencies on GitLab.  ItemRepository, the "client" folder with the "Git..." classes, and GitLabManager were the main dependencies to pull out. 

3. Move code out of common into their respective apps.


src/main/java/org/opentestsystem/ap/common/restclient/HistoryClient.java
```diff
@@ -0,0 +1,14 @@
+package org.opentestsystem.ap.common.restclient;
+
+import org.opentestsystem.ap.common.model.history.ItemHistory;
+import org.springframework.cloud.netflix.feign.FeignClient;
+import org.springframework.web.bind.annotation.PathVariable;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestMethod;
+
+@FeignClient(name = "historyClient", url = "${itembank.services.history.url}")
+public interface HistoryClient {
```
millerts: I feel like this belongs in history.  I wouldn't expect any endpoints to be in common.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
```diff
@@ -347,54 +344,37 @@ public ItemContext generateFiles(Item item, String bankKey, Path itemFolder) {
...
-        item.getCore().getEn().setPrompt("System generated Short Answer used to preview Stimulus");
-
-        String legacyId = item.getLegacyId(bankKey);
-
-        ItemContext itemContext = new ItemContext(itemId, legacyId, bankKey, this.appAssembler, repositoryPath);
-        transformerFactory.newTransformer(item.getType(), this.imageGenerator).generateSaaifPreview(item, bankKey, itemContext);
-
-        return createPreviewData(itemContext.getLocalRepositoryPath(), item, bankKey);
-    }
+//    /***
```
millerts: Can you add a TODO here explaining when this can be deleted or when it needs to change?

src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
```diff
@@ -0,0 +1,48 @@
...
+
+import java.util.Date;
+
+/**
+ * The public facing item history representation.  This is passed over the wire between services.
+ */
+@Data
+public class ItemHistory {
+
+    private String historyId;
```
millerts: Should these all have "" as the default value?

src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
```diff
@@ -0,0 +1,48 @@
...
+    private String itemCommitId;
+
+    private String branchCommitId;
+
+    private String contentVersionId;
+
+    private String commitBy;
+
+    @JsonFormat(pattern="yyyy-MM-dd'T'HH:mm:ss.SSS'Z'")
+    private Date commitDate;
```
millerts: Can we not use Instant on this rather than Date?

src/main/java/org/opentestsystem/ap/common/gitlab/GitLabPreviewManager.java
```diff
@@ -2,7 +2,7 @@
 
 import lombok.extern.slf4j.Slf4j;
 import org.opentestsystem.ap.common.itembank.IdGenerator;
-import org.opentestsystem.ap.common.model.PreviewData;
+import org.opentestsystem.ap.common.model.preview.PreviewData;
```
millerts: Is this a class that can be deleted?  Or renamed to not include the "Gitlab" in it?

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
```diff
@@ -136,74 +210,62 @@ public PreviewData generateAssociatedPreviewData(PreviewData previewData, List<S
...
-     * value in the generated data.
+     * Generates an item's preview data in the given folder.  The current version of the item is used.
      *
      * @param itemId     The item id.
      * @param bankKey    The bank key to use in the generated data.
      * @param itemFolder The folder to generate the data.
      * @return Information associated with the preview data generated.
      */
-    public PreviewData generatePreviewData(String itemId, String bankKey, Path itemFolder) {
+    private PreviewData _generatePreviewData(String itemId, String bankKey, Path itemFolder) {
```
millerts: Why put the `_` in front of the method?  I haven't seen that pattern used.

src/main/java/org/opentestsystem/ap/common/restclient/HistoryClient.java
bdrainer: This is the rest client for apps like IRS to use to communicate with history service.  It could go in IRS but I like it in common so apps that need it can configure Feign to look in the "restclient" common package.

src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
bdrainer: Yes

src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
bdrainer: I can look into that.  I did make it an Instant initially.  The json was interesting in the object is broken up into two properties.  I added the json format pattern and it broke it.  I'm sure there is way to do it, but I wanted to move on so I made it Date which is how it is most other places.

src/main/java/org/opentestsystem/ap/common/gitlab/GitLabPreviewManager.java
bdrainer: It is needed.  That is returned to IRS and IVS, so they have dependencies on it.  It is not GitLab specific.

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
bdrainer: I'll change that.  There were so many "generatePreviewData" methods I was getting confused.  I made all the private methods with an underscore.  PreviewManager came about because of some of this.  I moved duplicate code from DataStorePreviewManager and AsyncDataStorePreviewManager into PreviewManager.  That reduced the confusion.  No need for the underscore.

src/main/java/org/opentestsystem/ap/common/repository/ItemRepository.java
bdrainer: I removed the code.  It is not needed anymore.  I moved it out of ItemRepository and hope to have ItemRepository only used by the ItemBankSyncService (e.g. the only thing that talks to GitLab).

src/main/java/org/opentestsystem/ap/common/gitlab/GitLabPreviewManager.java
bdrainer: I'm sorry.  You meant GitLabPreviewManager.  I was seeing the diff and PreviewData.  My previous comment was about PreviewData.

GitLabPreviewManager can go away, but a little work is necessary for that.  Assuming migration does not need to talk to GitLab then we can move all GitLab/Git stuff into the ItemBankSyncService.

src/main/java/org/opentestsystem/ap/common/restclient/HistoryClient.java
alex-ponce: As you showed us in the review additional FeignClients will be added to this package:
`src/main/java/org/opentestsystem/ap/common/restclient`
I noticed the IRSConfiguration was component scanning this entire package. Does this mean that services that component scan this package will get an instance of all FeignClients even they might not use them?

src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
millerts: IMRT uses Instant.  If this is Jackson you need to use the JavaTimeModule to have it serialize/deserialize as you expect.

ISS Jackson Config.
```
            builder.featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
            builder.failOnUnknownProperties(false);
            builder.modules(new Jdk8Module(), new JavaTimeModule());
```

src/main/java/org/opentestsystem/ap/common/gitlab/GitLabPreviewManager.java
millerts: I guess this may depend on if the history tooling to get the older commits is going to be in migration.

PR #490, opened 2019-07-26
Feature/iat 3209

bdrainer:


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -132,10 +131,8 @@ public void updateAsl(String updateOwner, String updateOwnerFullName, ItemBulkUp
         asl.addAttachment(pair.getWebm().getFileName());
 
         // bulk ASL changes include updating the attachments zip
-        Optional<String> attachmentsVersionId = Optional.ofNullable(
-            this.attachmentManager.commitAttachments(
-                entityBranch.getItemJson(), entityBranch.getBranchName())
-        );
+        String attachmentsVersionId = this.attachmentManager.commitAttachments(
+            entityBranch.getItemJson(), entityBranch.getBranchName());
```
bdrainer: Attachment version ID is required when committing.  You'll see this change sprinkled in on this ticket.  It came up in IMS recently and I noticed during this ticket on bulk.

src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -346,7 +341,8 @@ public void patchItem(ItemBankUser user,
             user,
             patchedItem,
             message,
-            Optional.ofNullable(patchedItem.getServerManaged().getTimsAttachmentsVersionId()));
+            patchedItem.getServerManaged().getTimsAttachmentsVersionId()
```
bdrainer: There are no patch events in TIMS that changes attachments.  That means we don't need to create a new one for patch events but instead reuse the current attachments version.

src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -549,10 +544,8 @@ private void updateItemGlossary(ItemEntity entityMaster,
         });
 
         // bulk glossary changes include updating the attachments zip
-        Optional<String> attachmentsVersionId = Optional.ofNullable(
-            this.attachmentManager.commitAttachments(
-                entityBranch.getItemJson(), entityBranch.getBranchName())
-        );
+        String attachmentsVersionId = this.attachmentManager.commitAttachments(
```
millerts: Should these calls be in the underlying "commitItemAndSend" type methods.  Or is there a case where the commitItemAndSend doesn't fetch the entity branch's attachment version id?  I see this call in a few places which is why I ask.

src/main/java/org/opentestsystem/ap/common/datastore/entity/ItemEntity.java
```diff
@@ -54,12 +54,6 @@
     @Column(name = "is_deleted")
     private boolean deleted;
 
-    @Column(name = "created_by_fullname")
-    private String createdByFullName;
-
-    @Column(name = "commit_message")
-    private String commitMessage;
-
```
bdrainer: Removing this was really all that was required.  Our events already had the user fullname and the commit message.  The history service was changed to use those values where before it wasn't.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -364,12 +356,17 @@ public ItemEntity commitItem(ItemBankUser createdBy, ItemBankUser updatedBy, Ite
...
     public ItemEntity commitItem(Item branchItem,
                                  String branch,
-                                 Optional<String> attachmentsVersionId,
+                                 String attachmentsVersionId,
                                  ItemBankUser createdBy,
                                  ItemBankUser updatedBy,
                                  String commitMessage) {
         // never commit master on master, that is a patch
         if (!BRANCH_MASTER.equalsIgnoreCase(branch)) {
+            if(StringUtils.isBlank(attachmentsVersionId)) {
```
millerts: For example, should attachments version id be passed in or should this fetch it?  Based on my earlier comment above.

src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
bdrainer: Yes, not everything needs to commit attachments.  Patch events reuse the existing one.  

I'll take a closer look if there is a place to consolidate those places that do need to commit attachments.  Maybe a commitItemAndSend method that does not take attachment version ID but instead generates one (i.e. commits the attachments).

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
bdrainer: It makes sense, in this method, for attachmentsVersionId to be passed in.  It could be set on the branchItem and taken from there, so then it wouldn't be needed to be passed in.  We would have to think through if it makes sense to save attachmentVersions IDs on branch records.

This method is on the data-manager.  The item-manager called the attachment-manger to commit the attachment where it got back the attachmentsVersionId.  Item-manager is now calling data-manager to save the given data as a new master record in the database.  I wouldn't expect the data-manager to have any dependency on attachment-manager.

There might be places it makes sense to fetch it.  I'm open to that and will keep an eye out for it.

src/main/java/org/opentestsystem/ap/common/datastore/entity/ItemEntity.java
alex-ponce: Will the History service write the commitByFullName and the commitMessage for migrations commits? this was a miss when implementing the columns on the item table

bdrainer: Yes.  History receives events from the queue.  IMS, IBMS, and migration are responsible for putting the correct information in the event message which is user name, full name, and commit message.  It turns out they have been all along.  I choose not to include fullname in the history service because the item table had it but I realize now history has to have it or else importing the GitLab history would be missing that info.

PR #491, opened 2019-07-30
Remove the extra TI paragraphs

millerts:


src/main/resources/saaif_templates/mi-qti.ftl
```diff
@@ -12,7 +12,7 @@
     <#list columns as column>
     <#-- use character-based identifiers for columns -->
         <simpleAssociableChoice identifier="${chars[column?index]}">
-            <p style="width:7.0em">${column}</p>
```
millerts: Noticed no difference in the way it displays.

PR #492, opened 2019-07-30
Feature/iat 3181

bdrainer:
Fix for issue where tutorial/stim showed up when previewing a historic version of an item.


src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
```diff
@@ -119,43 +119,37 @@ public PreviewData generatePreviewData(ItemBankUser user, String itemId, boolean
...
+        String legacyId = this.getMasterItem(itemId).getItemJson().getLegacyId(bankKey);
 
-        // the item folder is located within a "root" folder
         Path itemFolder = this.generateRandomItemPath(legacyId);
 
         this.downloadHistoricContent(itemId, itemHistory.getContentVersionId(), itemFolder);
 
-        ItemContext itemContext = this.previewManager.newItemContext(item.getId(), legacyId, bankKey, itemFolder);
+        ItemContext itemContext = this.previewManager.newItemContext(itemId, legacyId, bankKey, itemFolder);
+        Item historicItem = itemContext.readModelFile();
```
bdrainer: This is the fix.  Read the item.json that belongs to the historic content.  Previously I was using the current master version of the item.

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
```diff
@@ -119,43 +119,37 @@ public PreviewData generatePreviewData(ItemBankUser user, String itemId, boolean
...
 
-        this.previewManager.fixBankKeyForHistoricPreview(itemContext, item, this.itemBankProperties.getBankKey());
+        this.previewManager.fixBankKeyForHistoricPreview(itemContext, historicItem, this.itemBankProperties.getBankKey());
 
-        if (item.getGlossary().getTerms().size() > 0) {
-            PreviewUtil.copyItemGlossaryForPreview(itemContext, item.getId(), bankKey);
+        if (Objects.nonNull(historicItem.getGlossary()) && isNotEmpty(historicItem.getGlossary().getTerms())) {
+            PreviewUtil.copyItemGlossaryForPreview(itemContext);
```
bdrainer: I think this change was unnecessary but fine to make as well.  It is unnecessary because the glossary property and terms property are never null.  We make sure of that in the code.  I wrote a test to confirm that confirms it.

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
```diff
@@ -119,43 +119,37 @@ public PreviewData generatePreviewData(ItemBankUser user, String itemId, boolean
...
         }
 
         itemContext.deleteGlossaryDirectory();
 
         // item is the latest version, preview data is set with the stim and tut it is associated with now,
         // not what it was historically
-        PreviewData previewData = PreviewUtil.newPreviewData(itemFolder, item, bankKey);
+        PreviewData previewData = PreviewUtil.newPreviewData(itemFolder, historicItem, bankKey);
```
bdrainer: Here is the other part of the fix.  Using the historic item as it has the correct associations to set in the PreviewData.

src/main/java/org/opentestsystem/ap/common/datastore/preview/PreviewManager.java
```diff
@@ -69,7 +69,7 @@ public PreviewData generatePreviewData(Item item, String branch, String bankKey,
         this.generateSaaif(itemContext, item, bankKey);
 
         if (item.getGlossary().getTerms().size() > 0) {
-            PreviewUtil.copyItemGlossaryForPreview(itemContext, item.getId(), bankKey);
+            PreviewUtil.copyItemGlossaryForPreview(itemContext);
```
bdrainer: I realized in some places I could use data on ItemContext instead of specific method arguments.

src/test/resources/model/item-bare-min.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "sa"
+}
```
bdrainer: I have an idea where we can put as many variations of item json files.  Then have tests to confirm expectations.  Maybe an item json per item type and have a test suite verify we have not change serializing/deserializing.  Maybe deserialize the file, serialize back to file, then diff the two files expecting no differences.

I wonder if this is a good idea for SAAIF as well.  Have a folder "saaif" with as many xml variations we want, metdata included.  There would be test suites to ensure we haven't changed anything unexpected.  

I could even see having zips in the test resources to confirm nothing changed like if an item is marked not managed by IAT we would expect not to overwrite the given QRX and GAX.  The zip would contain exactly what we expect for that given test.

bdrainer:
I am going to merge so I can pin this and get the fix into QA.  Please review, any feedback changes will be applied in a new PR.

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
```diff
@@ -119,43 +119,37 @@ public PreviewData generatePreviewData(ItemBankUser user, String itemId, boolean
...
-        ItemContext itemContext = this.previewManager.newItemContext(item.getId(), legacyId, bankKey, itemFolder);
+        ItemContext itemContext = this.previewManager.newItemContext(itemId, legacyId, bankKey, itemFolder);
+        Item historicItem = itemContext.readModelFile();
 
-        this.previewManager.fixBankKeyForHistoricPreview(itemContext, item, this.itemBankProperties.getBankKey());
+        this.previewManager.fixBankKeyForHistoricPreview(itemContext, historicItem, this.itemBankProperties.getBankKey());
 
-        if (item.getGlossary().getTerms().size() > 0) {
-            PreviewUtil.copyItemGlossaryForPreview(itemContext, item.getId(), bankKey);
+        if (Objects.nonNull(historicItem.getGlossary()) && isNotEmpty(historicItem.getGlossary().getTerms())) {
```
millerts: Is there ever an instance when items don't have a glossary?

src/main/java/org/opentestsystem/ap/common/datastore/preview/DataStorePreviewManager.java
millerts: You can ignore my comment above.  I agree it is unnecessary as it indicates a bug in the model.

PR #494, opened 2019-08-01
IAT-3210 - Add diff to the history service

alex-ponce:
- Added `contentDiff` property to `ItemHistory` model
- Added `findLatest` method to `ItemHistoryRepository`
- Added `DiffUtil` class, it expose a method to generate Diff content based on two directories
- Added unit tests


src/main/java/org/opentestsystem/ap/common/util/DiffUtil.java
```diff
@@ -0,0 +1,104 @@
...
+     * @param oldCommit      RevCommit of previous (old) commit
+     * @param newCommit      RevCommit of new commit
+     * @return               Diff syntax content
+     * @throws IOException   IOException
+     */
+    private static String generateDiff(Git git, RevCommit oldCommit, RevCommit newCommit) throws IOException {
+        AbstractTreeIterator oldTreeIterator = getCanonicalTreeParser(git, oldCommit);
+        AbstractTreeIterator newTreeIterator = getCanonicalTreeParser(git, newCommit);
+
+        ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
```
alex-ponce: The JGit Api uses an output stream to capture the Diff Content it generates

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -430,4 +431,10 @@ public ItemSyncEntity saveItemSync(ItemSyncEntity itemSync) {
     public ItemHistoryEntity saveItemHistory(ItemHistoryEntity itemHistory) {
         return this.itemHistoryRepository.save(itemHistory);
     }
+
+    @Transactional
+    public Optional<ItemHistoryEntity> findLatestItemHistory(String itemId) {
+        ItemHistoryEntity latest = this.itemHistoryRepository.findLatest(itemId);
+        return latest == null ? Optional.empty() : Optional.of(latest);
```
bdrainer: No need to change this but you can do just Optional.ofNullable(latest).  If you look at the code for the 'ofNullable" method it does exactly what you are doing there

src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemHistoryRepository.java
```diff
@@ -13,4 +15,10 @@
     @Query("SELECT entity FROM ItemHistoryEntity entity WHERE entity.itemId = :itemId ORDER BY entity.commitDate DESC")
     List<ItemHistoryEntity> findByItemId(@Param("itemId") String itemId);
 
+    @Query("SELECT entity FROM ItemHistoryEntity entity WHERE entity.itemId = :itemId"
+        + " AND entity.commitDate = "
+        + "     (SELECT MAX(d.commitDate) FROM ItemHistoryEntity d WHERE d.itemId = :itemId )"
+    )
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    ItemHistoryEntity findLatest(@Param("itemId") String itemId);
```
bdrainer: Is this supposed to be the "lock" query?  I see it is getting the latest history record and locking on it.  If you need to lock I would lock on the MIN(d.commitDate).  

Maybe consider having two methods: one to lock method  and then one to get the latest history record for an item.

What do you need the latest history record for an item for, is it so you can set the diff on it?

bdrainer: I remember why we need to lock.  We need to lock so we make sure we get the current latest record to compare against.

I would go with two methods.  One to lock on the oldest record.  I feel the oldest record is better to lock on rather than the current latest record.  The oldest record never changes but the current latest does.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
alex-ponce: I like less code!

src/main/java/org/opentestsystem/ap/common/util/DiffUtil.java
```diff
@@ -0,0 +1,104 @@
...
+            git.add().addFilepattern(".").call();
+            RevCommit newCommit = git.commit().setMessage("New Commit").call();
+
+            diff = generateDiff(git, oldCommit, newCommit);
+
+            git.getRepository().close();
+
+        } catch (Exception e) {
+            log.error("Unable to generate diff content", e);
+            throw new SystemException(e);
```
millerts: Is this double logging?  As in we log here and then the upstream system also logs?  I recommend adding a message to the SystemException with the `e` and not logging here knowing the upstream code will log.

src/main/java/org/opentestsystem/ap/common/util/DiffUtil.java
```diff
@@ -0,0 +1,104 @@
...
+            gitRepo = Files.createTempDirectory("gitRepo_");
+            Git git = Git.init().setDirectory(gitRepo.toFile()).call();
+
+            // Copy, add and commit old files
+            FileUtils.copyDirectory(oldCommitDir.toFile(), gitRepo.toFile());
+            git.add().addFilepattern(".").call();
+            RevCommit oldCommit = git.commit().setMessage("Old Commit").call();
+
+            // Copy, add and commit new files
+            FileUtils.copyDirectory(newCommitDir.toFile(), gitRepo.toFile());
```
millerts: Can you add a finally to this try/catch that cleans up all these directories?  Or is that handled somewhere else?

src/main/java/org/opentestsystem/ap/common/util/DiffUtil.java
alex-ponce: Removed try/catch and left it up to the caller to handle it

src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemHistoryRepository.java
alex-ponce: Added proper lock method to ItemHistoryRepository

src/main/java/org/opentestsystem/ap/common/util/DiffUtil.java
alex-ponce: Added directory cleanup

src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemHistoryRepository.java
alex-ponce: Added proper lock method

PR #500, opened 2019-08-22
Feature/iat 3130

bdrainer:
We are adding a column to the ITEM_HISTORY table called gitlab_commit_hash.  This is the same name we used in the ITEM_SYNC table.

The changes in the PR are around getting item history records where gitlab_commit_hash is not null.  Those records identify what history records were imported from GitLab.

Most the changes are in the ItemHistoryImporter project.  PR coming shortly for that.


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -203,17 +203,45 @@ public void downloadAttachmentsForPreview(String itemId,
...
+    /**
+     * The zip file is expected to be a zip file and not a folder that is expected to be zipped.
+     *
+     * @param itemId  The item id.
+     * @param zipFile The file to write to the data store.
+     * @return The content version id.
+     */
+    public String commitSaaifContentZip(String itemId, Path zipFile) {
+        String zipName = this.getSaaifContentZipName(itemId);
+        return this.s3VersionedBucketManager.uploadFile(zipName, zipFile);
```
bdrainer: GitLab has an API to download a specific commit as a zip.  Pretty cool.  We download that archive zip then stream it up to S3 with the name we want to use.

src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemHistoryRepository.java
```diff
@@ -17,14 +17,41 @@
...
+     *
+     * @param itemId The item id.
+     * @return Imported history records with no content version id.
+     */
+    @Query("SELECT entity FROM ItemHistoryEntity entity WHERE entity.itemId = :itemId"
+        + " AND entity.gitCommitHash <> null"
+        + " AND entity.contentVersionId is null"
+        + " ORDER BY entity.commitDate asc"
+    )
+    List<ItemHistoryEntity> findImportedRecordsWithNullContentVersion(@Param("itemId") String itemId);
```
bdrainer: The importer uses this to find the history records that need content version created and saved to S3.

src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemHistoryRepository.java
```diff
@@ -17,14 +17,41 @@
...
+    /**
+     * Used by the GitLab commit history import process.  Imported records have a value for the 'gitCommitHash' property
+     * where non-imported have null.  This method returns the imported records ordered from oldest to newest.
+     *
+     * @param itemId The item id.
+     * @return Imported history records with no content version id.
+     */
+    @Query("SELECT entity FROM ItemHistoryEntity entity WHERE entity.itemId = :itemId"
+        + " AND entity.gitCommitHash <> null"
+        + " ORDER BY entity.commitDate asc"
```
bdrainer: This query is used by the importer to get a list of the GitLab imported records.  We skip the first (oldest) commit as we don't need its diff.  Once we get to the second item in the list we start generating diff.  The second record in the list is diff'ed against the first record.  The third record is diff'ed against the second record. Etc. Etc.

PR #503, opened 2019-09-03
IAT-3248 - Translation service should only include SAAIF referenced files

alex-ponce:
- Added `ContentJobParams.includeTutorial` property
- Implemented `DataStoreAttachmentManager.downloadReferencedAttachments` and DataStoreAttachmentManager.downloadReferencedGlossaryAttachments`. These methods download from S3 only the items referenced items
- Created `AttachmentSelector` class that walks the Item model and returns a list of filenames currently referenced on the item.
- Added integration tests


src/main/java/org/opentestsystem/ap/common/batch/ContentJobParams.java
```diff
@@ -27,6 +27,8 @@
 
     private boolean includeImport;
 
+    private boolean includeTutorial;
+
```
alex-ponce: This property is used to save the input parameters of the Saaif Content Batch job

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -402,4 +435,27 @@ private String getAttachmentsZipName(final String itemId) {
...
+        fileNames.forEach(fileName -> {
+            Path outputPath = rootPath.resolve(fileName);
+
+            String fileKey = String.format("%s/%s", fileKeyPrefix, fileName);
+
+            try (InputStream inputStream = s3BucketManager.downloadFile(fileKey, null)) {
+                writeFile(outputPath, inputStream);
+            } catch (Exception e) {
+                log.error(String.format("[FILE-NOT-FOUND-IN-S3] Unable to download file using keyFile: %s", fileKey), e);
+            }
```
alex-ponce: Implemented error handling to log missing files rather than throw. This will allow the job to continue. This scenario is rare but occurred on a test item.

src/test/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManagerTest.java
```diff
@@ -0,0 +1,100 @@
...
+import java.io.InputStream;
+import java.nio.file.Files;
+import java.nio.file.Path;
+
+/**
+ * These integration tests require the existence of several item attachment buckets in S3
+ * These tests help during development of DataStoreAttachmentManager
+ * All tests are ignored to avoid issues in CI builds
+ */
+@Slf4j
```
alex-ponce: As comment above indicates, these integration tests are ignored and only used during development

PR #512, opened 2019-10-02
Feature/iat 3185

bdrainer:
Todd you have reviewed some of this already.  It was a PR on the permission service.  I decided to move everything to common, specifically to share the authenticated user model that gets stored in the session cache. 

Most this PR is the admin user/role/permission stuff, and some is the security util pulling the user out of the security context.


src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+
+        for (UserRoleEntity userRole : roles) {
+            for (RolePermissionEntity rolePermission : userRole.getRole().getPermissions()) {
+                this.mapPermissions(permissionMap, rolePermission);
+            }
+        }
+
+        List<PermissionModel> userPermissions = permissionMap.entrySet().stream()
+            .map(e -> AdminMapper.mapToPermissionModel(e.getKey(), e.getValue()))
+            .collect(Collectors.toList());
```
bdrainer: This is where we are taking the user's roles (UserRole) and flatten the permissions associated with it.  We are returning a user with a list of permissions.  The permissions in the list might have a list of workflow statuses with it.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+        RoleEntity roleEntity = this.dataManager.findRole(roleId).orElseThrow(roleNotFound(roleId));
+
+        Set<RolePermissionEntity> rolePermissions = roleEntity.getPermissions();
+
+        // this flattens the permissions and workflow statuses
+        rolePermissions.stream().forEach(rolePermission -> this.mapPermissions(permissionMap, rolePermission));
+
+        List<PermissionModel> userPermissions = permissionMap.entrySet().stream()
+            .map(e -> AdminMapper.mapToPermissionModel(e.getKey(), e.getValue()))
+            .collect(Collectors.toList());
```
bdrainer: Similar to user when we get a Role we flatten the permissions associated with it, along with their associated workflow statuses.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+        } else {
+            permissionMap.put(permission, new HashSet<>(workflowStatusCodes));
+        }
+    }
+
+    @Transactional
+    public RoleModel createRole(String roleLabel) {
+        this.validator.validateRoleCreate(roleLabel);
+        return AdminMapper.mapToRoleModel(this.dataManager.createRole(roleLabel, this.getAuthenticatedUser().getId()));
+    }
```
bdrainer: We discussed creating a role only takes a label value.  We are not allowing the permission associations to be part of role creation.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+                    .collect(Collectors.toSet());
+
+            })
+            .flatMap(Set::stream)
+            .collect(Collectors.toSet());
+
+        this.dataManager.saveRolePermissionWorkflowStatuses(rolePermissionWorkflowStatuses);
+
+        return this.findRole(roleId);
+    }
```
bdrainer: Updating role was the hardest part.  It was hard because the relationship was difficult to do in Hibernate, particularly the role to permission to workflow status.  Role and permission have a join table Role_Permission, and that join table creates the join relationship with workflow status thus having another join table.  I tried what I understood with Hibernate but couldn't get it to work how I wanted.

My solution was to use Hibernate for all this but manage the RolePermissionWorkflow status independently meaning there are no relationships defined on it.  Doing it this way means we have to be aware of flushing.  We have to delete RolePermissionWorkflow before deleting the RolePermission relationship.  The opposite for insert, for example when associating a permission to a role we have to insert to RolePermission first, flush, then insert to RolePermissionWorkflowStatus.

The above method is simplistic and will have to change when we start auditing changes.  Currently when a role is being updated I'm deleting everything associated with that role from both RolePermissionWorkflowStatus and RolePermission.  Then I insert everything that is in the request.  

There are a few things to be aware with this approach.  It will look like all role to permission relationships were created during an update however, if a permission(s) was already associated with the role we lost when that first happened.  I think this is something we can address when we introduce auditing.  The idea would be find out what permissions are being deleted, what permissions are being added, and what permissions are already there and not changing.  Then delete only those that should be and add only those that should, leaving the ones not changing untouched.

src/main/java/org/opentestsystem/ap/common/admin/api/AdminPermissionsApi.java
```diff
@@ -0,0 +1,43 @@
...
+import org.springframework.web.bind.annotation.RestController;
+
+import java.util.List;
+
+/**
+ * Admin Permissions Api
+ */
+@Slf4j
+@RestController
+@RequestMapping("/api/v1/admin/permissions")
```
millerts: Why is the API in common?  I know at some point we want to have multi module common, but I'm used to the Web API's still remaining in the services.

millerts: ^ Any of the HTTP API's I'm seeing in common

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+
+    // ------------------------------------------------------------------------
+    // Organization Type
+    // ------------------------------------------------------------------------
+
+    public List<OrganizationTypeModel> findAllOrganizationTypes() {
+        return this.dataManager.findOrganizationTypes().stream().map(OrganizationTypeModel::map)
+            .collect(Collectors.toList());
+    }
```
bdrainer: I almost feel like this should be somewhere else.  WorkflowStatus is served by IMS.  The bigger picture is the idea of "content" where we have a content manager feeding things like this to the TIMS UI

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
```diff
@@ -0,0 +1,178 @@
...
+import static org.apache.commons.lang3.StringUtils.trimToEmpty;
+import static org.apache.commons.lang3.math.NumberUtils.INTEGER_ONE;
+
+@Component
+public class AdminMapper {
+
+    public static UserEntity mapToNewUserEntity(ChangeUserRequest model,
+                                                Integer createdBy,
+                                                Collection<RoleEntity> roles) {
+        UserEntity entity = AdminMapper.mapToUserEntity(model);
```
millerts: Should the AdminMapper.mapToUserEntity take in the version and createdBy?  Since it always seems required?

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
```diff
@@ -0,0 +1,178 @@
...
+        entity.setRoles(mapToUserRoleEntity(entity.getId(), createdBy, roles));
+        entity.setVersion(INTEGER_ONE);
+        entity.setCreatedBy(createdBy);
+        entity.setUpdatedBy(createdBy);
```
millerts: Probably too late but we should remove the either createdBy or updatedBy column since they are **always** the same do to the always insert policy.

src/main/java/org/opentestsystem/ap/common/security/SecurityUtil.java
```diff
@@ -47,8 +61,14 @@
 
     public static final String USER_ANONYMOUS = "anonymous";
 
+     public AuthenticatedUserModel getAuthenticatedUser() {
+        return this.getOptionalAuthenticatedUser().orElse(this.anonymousAuthenticatedUser());
+    }
```
bdrainer: The "user" instance in the security context now has a reference to AuthenticatedUserModel.  If this is null on the user then the user is not authorized for anything.  We are going to use the model on the user for everything permission check related.  So if we get null for the model we know to deny permission.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
```diff
@@ -0,0 +1,178 @@
...
+import java.util.Set;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+import static java.util.Collections.emptyList;
+import static org.apache.commons.collections4.CollectionUtils.isNotEmpty;
+import static org.apache.commons.lang3.StringUtils.trimToEmpty;
+import static org.apache.commons.lang3.math.NumberUtils.INTEGER_ONE;
+
+@Component
```
millerts: Does this need to be a Component?  It looks like everything in this class is static.  Recommend either making the methods non static or removing component.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
```diff
@@ -0,0 +1,178 @@
...
+public class AdminMapper {
```
millerts: So I'm following the pattern any mapping in the Admin area of the application (going from one object to another) should be located in this class?

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
millerts: Why have this mapping here rather than have all the mapping be in AdminMapper.mapToAuthenticatedModel?

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+        List<PermissionModel> userPermissions = permissionMap.entrySet().stream()
+            .map(e -> AdminMapper.mapToPermissionModel(e.getKey(), e.getValue()))
+            .collect(Collectors.toList());
+
+        return AdminMapper.mapToAuthenticatedModel(user, userPermissions);
+    }
+
+    @Transactional(readOnly = true)
+    public List<UserModel> findAllUsers() {
+        log.info("Find all users");
```
millerts: Can we make the entry type messages debug or remove them?

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
```diff
@@ -0,0 +1,331 @@
...
+    @Transactional(readOnly = true)
+    public UserModel findUser(Integer id) {
+        log.info("Find user {}", id);
+        UserEntity user = this.dataManager.findUser(id).orElseThrow(userNotFound(id));
+        return AdminMapper.mapToUserModel(user);
+    }
+
+    @Transactional
+    public void deleteUsers(List<Integer> usersToDelete) {
+        log.info("Delete users {}", usersToDelete);
```
millerts: I would leave this at info.

millerts: Actually, I defer to you on that since all of this information is in the db.

src/main/java/org/opentestsystem/ap/common/admin/api/AdminPermissionsApi.java
bdrainer: I'll move them back to tims-perm-service.  It moved it thinking one day we will have common be multi-module so I'll stick the whole admin module in there now.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
millerts: I have the same question about why this is here vs the mapper.  The conversion of data to another projection.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
bdrainer: We aren't doing insert only anymore.  There was a discussion and we are updating now, so updateBy on user, role, and permission makes sense.  Though I suspect audit will be what they use to see who did what when.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminValidator.java
```diff
@@ -0,0 +1,256 @@
...
+@Component
+public class AdminValidator {
+
+    private final AdminDataManager adminDataManager;
+
+    public AdminValidator(AdminDataManager adminDataManager) {
+        this.adminDataManager = adminDataManager;
+    }
+
+    public static Supplier<ResourceNotFoundException> userNotFound(Integer id) {
```
millerts: These can all be private, right?  I didn't check all of them but they seemed to be isolated to this class?

src/main/java/org/opentestsystem/ap/common/management/ItemManager.java
```diff
@@ -45,14 +45,18 @@
 
     private final EquationEditorRepository equationEditorRepository;
 
+    private final SecurityUtil securityUtil;
+
     public ItemManager(DataStoreItemManager dataStoreManager,
                        GitLabItemManager gitLabManager,
```
millerts: Not related to your changes but I'd love to see this ripped out if possible.

src/main/java/org/opentestsystem/ap/common/model/OrganizationTypeModel.java
```diff
@@ -0,0 +1,31 @@
...
+@Builder
+public class OrganizationTypeModel {
+
+    private Integer id;
+
+    private String code;
+
+    private String label;
+
+    public static OrganizationTypeModel map(OrganizationTypeEntity organizationTypeEntity) {
```
millerts: Shouldn't this be in AdminMapper to follow the pattern?

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
bdrainer: Yes.  Admin entity to/from admin model.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
bdrainer: yes

src/main/java/org/opentestsystem/ap/common/admin/service/AdminValidator.java
bdrainer: The suppliers are used in the service class.  For example

`UserEntity user = this.dataManager.findUser(email).orElseThrow(AdminValidator.userNotFound(email));`

src/main/java/org/opentestsystem/ap/common/security/SecurityUtil.java
```diff
@@ -170,59 +196,64 @@ public static boolean isAssigner(ItemBankUser user, Item item) {
...
-        return fullname;
+        return Optional.empty();
     }
 
     @VisibleForTesting
     Authentication getAuthentication() {
         return SecurityContextHolder.getContext().getAuthentication();
     }
+
+    private User anonymousUser() {
```
millerts: Is this just for local development?  I'm curious if we ever need an anonymous user really.

src/main/java/org/opentestsystem/ap/common/model/OrganizationTypeModel.java
bdrainer: Yes it should

src/main/java/org/opentestsystem/ap/common/admin/api/AdminPermissionsApi.java
millerts: We can leave it in here for now if needed since we know we have other tech debt that needs to be addressed.  My main concern is I would like patterns that one can follow or reason with without having to reach out to another developer.  Your call.

src/main/java/org/opentestsystem/ap/common/security/SecurityUtil.java
bdrainer: This is how we have always had it but we can reject things if they are anonymous.  To elaborate when calling into IMS, IRS, etc via Postman then asking the Spring security context for the user, it would return a UserDetail instance (or maybe just a string) for 'anonymousUser'.  Spring was doing that by default, so we gave it no roles or permissions.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
bdrainer: No it doesn't.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminMapper.java
bdrainer: Since this is specifically for "new" user it seemed like a good idea to let the mapper set the  "first version" value TIMS uses which is 1.  However it can just as easily be passed in and we will use the value 1 when doing it.

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
bdrainer: It's a good point.  There is a call in this method `this.mapPermissions(permissionMap, rolePermission);` that looks up in the database the workflow statuses associated with the RolePermission.  If it wasn't for this I would move it to the mapper.  

Ideally in a Hibernate world we could look up User and ask for its properties and let Hibernate under the covers get everything for us.  For the most part that is true except for the pesky  RolePermissionWorkflowStatus.

src/main/java/org/opentestsystem/ap/common/security/SecurityUtil.java
bdrainer: I think there is a flag/config setting to tell Spring not to allow anonymous.  I'll look into that

src/main/java/org/opentestsystem/ap/common/admin/service/AdminService.java
bdrainer: Same reason, the RolePermissionWorkflowStatus is looked up prior to calling the mapper.  So the service is gathering the data from the DB then giving it to the mapper.   Again, if only I could get the Hibernate relationship mapped correctly for RolePermissionWorkflow status then we could rely on Hibernate load the properties when we ask for them.

src/main/java/org/opentestsystem/ap/common/admin/api/AdminPermissionsApi.java
bdrainer: It was easy enough to move. and more like move back as it started there originally.

PR #513, opened 2019-10-02
Bugfix/iat 3285

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -199,25 +199,6 @@ public Long deleteItem(String itemId) {
...
-     * <p>
-     * Insert differs from save. Save expects changes to come from a user editing a branch or when creating the item.
-     * Save first looks up the branch the user is editing so it knows what to make the save against.  Insert simply
-     * inserts the item entity.  Everything on the item json should be set, so the version should already be set.
-     *
-     * @param entity
-     * @return
-     */
-    @Transactional
-    public ItemEntity insertItem(ItemEntity entity) {
```
millerts: This was only used by item migration.  Spoke with @bdrainer , and we both agree that direct calls to item related methods in this class should be discouraged and all item related calls should go through DataStoreItemManager.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -166,6 +168,64 @@ public void systemCommitItem(String userName, Item item, String commitMessage) {
...
+     * Saves a branch migrated item.
+     *
+     * @param systemUser the {@link ItemBankSystemUser} to save the item
+     * @param entity     the migrated {@link ItemEntity}
+     * @return the saved entity
+     * @throws IllegalArgumentException if the entity is the master branch
+     */
+    @Transactional
+    public ItemEntity saveMigratedBranchItem(final ItemBankSystemUser systemUser, final ItemEntity entity) {
+        if (entity.isMaster()) {
```
millerts: The logic around branch migrations is slightly different than master migrations.  I wanted to use the existing methods but migrations are just different enough from the normal flow that I would have had to do a pretty big refactor to make the existing item modificatin logic flexible enough for migraiton.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -166,6 +168,64 @@ public void systemCommitItem(String userName, Item item, String commitMessage) {
...
+        this.dataManager.lockMaster(entity.getItemId());
+
+        final Item item = entity.getItemJson();
+
+        dataStoreUtility.updateServerManagedFields(item);
+        dataStoreUtility.updateRubric(item);
+
+        boolean isMasterBranch = BRANCH_MASTER.equalsIgnoreCase(entity.getBranchName());
+
+        if (!isVersionOnlyUpdate && isMasterBranch) {
```
millerts: We need to updateMasterMedataVersion whenever migrations modify the item.

PR #516, opened 2019-10-07
Feature/iat 3185

bdrainer:
Some minor changes to common related to the admin feature.


src/main/java/org/opentestsystem/ap/common/datastore/entity/admin/PermissionCategoryEntity.java
```diff
@@ -30,7 +31,7 @@
     private Integer order;
 
     @OneToMany(mappedBy = "category", cascade = CascadeType.ALL, orphanRemoval = true)
-    private Set<PermissionEntity> permissions;
+    private Set<PermissionEntity> permissions = new HashSet<>(0);
```
bdrainer: Need to default this to non-null.

src/main/java/org/opentestsystem/ap/common/datastore/entity/admin/RoleEntity.java
```diff
@@ -43,15 +43,15 @@
     @Version
     protected Integer version;
 
+    @OneToMany(mappedBy = "role", cascade = CascadeType.ALL, orphanRemoval = true)
+    private Set<RolePermissionEntity> permissions = new HashSet<>(0);
+
```
bdrainer: A simple move of this code piece above its constructor.

src/main/java/org/opentestsystem/ap/common/rest/ExceptionAdvice.java
```diff
@@ -56,6 +58,12 @@
 @Order(Ordered.HIGHEST_PRECEDENCE)
 public class ExceptionAdvice extends ResponseEntityExceptionHandler {
 
+    @ExceptionHandler(UnauthorizedException.class)
+    public ResponseEntity<SimpleError> handleUnauthorized(final UnauthorizedException ex) {
+        log.warn("{}", ex.getMessage());
+        return new ResponseEntity<>(new SimpleError(ex.getMessage()), jsonHeaders(), UNAUTHORIZED);
+    }
+
```
bdrainer: Added this as we will likely be throwing unauthorized from different apps now.  There is no where yet we are throwing this and maybe we won't need this.  We might also need an advice handler for Not Allowed.

millerts: It should probably throw if we ever add validation to IMS.  If you are trying to access an item that is a type you can't view one should get a not authorized 401 (I think that is the code?).

src/main/java/org/opentestsystem/ap/common/security/SecurityUtil.java
```diff
@@ -212,6 +215,14 @@ private User getUserOrAnonymous() {
...
             if (principal instanceof User) {
                 return Optional.of((User) principal);
+            } else if (principal instanceof String) {
+                return Optional.of(User.builder().username((String) principal).build());
+            } else if (principal instanceof UserDetails) {
+                UserDetails details = (UserDetails) principal;
+                return Optional.of(
+                    User.builder()
+                        .username(details.getUsername())
+                        .build());
```
bdrainer: I put back how we used to have this.  We used to look for User and if not found then look for either UserDetails or String from the principal.  This is here in the case someone has an authenticated user in the security context, meaning the principal user was auth'ed but there maybe not be an AuthenticatedUserModel in the security context.

PR #517, opened 2019-10-16
Feature/iat 3325

millerts:


src/main/java/org/opentestsystem/ap/common/model/ModelConstants.java
```diff
@@ -135,6 +138,22 @@
         String TYPE_HTQS = "htqs";
         String TYPE_GI = "gi";
         String TYPE_WORD_LIST = "wordList";
+
+        Set<String> TIMS_ITEM_TYPES = Sets.newHashSet(
```
millerts: These are the item types that are supported by TIMS.  Occasionally I need to iterate over all the item types so I created an Set.

src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
```diff
@@ -373,11 +394,11 @@ public boolean exportItemImages(String workflowStatusCode) {
     }
 
     public boolean previewItemAllVersions(String workflowStatusCode) {
-        return this.has("preview_item_all_versions", workflowStatusCode);
+        return this.has(previewAllItemVersions, workflowStatusCode);
```
millerts: I made these constants because the way IMRT needs to deal with permissions can't leverage this capabilities class.

src/main/java/org/opentestsystem/ap/common/model/admin/PermissionCode.java
```diff
@@ -0,0 +1,37 @@
+package org.opentestsystem.ap.common.model.admin;
+
+/**
+ * The permissions codes in the application
+ */
+public class PermissionCode {
+
+    public static final String accessTims = "access_tims";
```
millerts: I only moved the permission codes that are needed by IMRT.

PR #519, opened 2019-10-16
Feature/iat 3306

jeffjohnson9046:
[IAT-3306](https://smarterbalanced.atlassian.net/browse/IAT-3306):  Rich Text Editor for TIMS Short Answer Items

* Add `textEditorType` to `SaItemContent`
* Add `TextEditorTypeEnums` to describe the available text editor types
* Update `getResponseType` to return the appropriate SAAIF attribute value based on the selected text editor


src/main/java/org/opentestsystem/ap/common/model/content/TextEditorTypeEnums.java
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Describe the available text editor input fields for Short Answer (SA) items.
+ */
+@Getter
+public enum TextEditorTypeEnums {
+    /**
+     * No text editor selected.
+     */
+    Empty(EMPTY),
```
millerts: Does Jackson know to set the enum to empty if it is an empty string?  I only ask because you define the properties for the other ones even though enum is the exact match for the property.

src/main/java/org/opentestsystem/ap/common/model/SaItem.java
```diff
@@ -44,9 +42,17 @@ public String getMetadataType() {
...
+    /**
+     * @return The response type based on the type of text editor selected for this item.
+     * <p>
+     * The type of text editor is consistent across translations; e.g. the text editor students can use should be
+     * the same if the item is in English and has a Spanish translation.  As such looking at the "EN" translation is
+     * sufficient.
+     * </p>
+     */
     @Override
     public String getResponseType() {
```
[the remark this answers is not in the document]
jeffjohnson9046: Unfortunately, I'm not sure.  I left the `@Override` in place here to be as "surgical" as possible with my change.

src/main/java/org/opentestsystem/ap/common/model/content/TextEditorTypeEnums.java
jeffjohnson9046: I've made two changes:

* removed the `@JsonProperty` from the individual enums and added `@JsonValue` to the `getSaaifResponseAttributeValue()` method
* Changed the `Empty` enum to `NoneSelected`, which is more semantic in my opinion

PR #522, opened 2019-10-18
Feature/iat 3327

bdrainer:
These changes are for the server side permission checks. It is similar to the front end checks.  The ItemBankUser is constructed from what is in the security context however, the async bulk processing for ASL, braille, and glossary required looking up the user's permissions from the TIMS permission service.


bdrainer:
Once these common changes get in develop there are associated changes to get in for IAT and IBMS.

src/main/java/org/opentestsystem/ap/common/bulkupdate/processor/ItemBulkUpdateProcessor.java
```diff
@@ -87,6 +94,11 @@ protected ItemBulkUpdateProcessor(Path root,
 
     // ------------------------------------------------------------------------
 
+    protected ItemBankUser lookupItemBankUser(String username) {
```
millerts: I don't know this area of code but can this be called at root level?  Or is this the root level of the flow?  I thought there was a higher level class that could pass the authenticated user to the processors.

src/main/java/org/opentestsystem/ap/common/security/model/UserBuilder.java
```diff
@@ -0,0 +1,109 @@
+package org.opentestsystem.ap.common.security.model;
+
+import org.opentestsystem.ap.common.model.admin.AuthenticatedUserModel;
+import org.springframework.security.core.GrantedAuthority;
+
+import java.util.Collection;
+
+public class UserBuilder {
```
millerts: Not sure where the User class exists but you could have Lombok make a builder for it with the `@Builder` annotation.  I think it would be similar to this class.

src/main/java/org/opentestsystem/ap/common/bulkupdate/processor/ItemBulkUpdateProcessor.java
bdrainer: Yes.  That is a good idea.  It would be the ItemBulkUpdateEventConsumer.  This class currently gets the message off the queue, queries the DB for the bulk update record and hands that off to the processors.  It will now look up the user from the permission service and hand it off to the processors as well.

src/main/java/org/opentestsystem/ap/common/security/model/UserBuilder.java
bdrainer: It's a good point and actually that is how we had it.  I kept getting an odd serialization error as I was making these changes and the one I kept seeing was when serializing User it would complain about the Lombok UserBuilder.  This likely isn't an issue with the Lombok builder given that is how we always had and it was never an issue, but to rule it out I rolled my own builder.

PR #524, opened 2019-10-24
Added 'findActive' RoleReposity method

alex-ponce:
Added 'findActive' RoleReposity method to filter out deleted roles.


src/main/java/org/opentestsystem/ap/common/datastore/repository/admin/RoleRepository.java
```diff
@@ -16,6 +18,9 @@
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     RoleEntity lock(@Param("id") Integer id);
 
+    @Query("SELECT entity FROM RoleEntity entity WHERE entity.deleted = false")
+    List<RoleEntity> findActive(Sort sort);
+
```
tims-fairway-system: Nice.  Where is this being  called?

PR #528, opened 2019-11-01
Feature/iat 3346

millerts:


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -818,7 +818,7 @@ private void updateGlossaryImages(ItemEntity entityMaster,
             }
         }
 
-        itemManager.commitItem(this.newUser(updateOwner, updateOwnerFullName), item, commitMessage);
+        itemManager.commitItem(this.newUser(updateOwner, updateOwnerFullName), systemUser(), item, commitMessage);
```
millerts: This was the bug I mentioned in the channel.  I may come back through this area at a later date and do some light refactoring consolidating a lot of the duplicated code in this class.

PR #529, opened 2019-11-01
Add logic around migration for the new attachment object

millerts:
Fixes migration specific code.


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -208,14 +209,21 @@ public ItemEntity saveMigratedBranchItem(final ItemBankSystemUser systemUser, fi
...
+                                       final String commitMessage,
+                                       final Optional<AttachmentCommitResult> attachmentCommitResult) {
         this.dataManager.lockMaster(entity.getItemId());
 
         final Item item = entity.getItemJson();
 
         dataStoreUtility.updateServerManagedFields(item);
         dataStoreUtility.updateRubric(item);
 
+        attachmentCommitResult.ifPresent(attachmentCommitResult1 -> {
```
millerts: Need to process the attachment commit results

PR #530, opened 2019-11-02
Feature/iat 3350

bdrainer:
I tested this using the itembank sync service.  It required changing three lines in the ItemManagerEventHandler.  I doubt preview or validation will be that easy but here's to hoping.


src/main/java/org/opentestsystem/ap/common/model/Attachment.java
```diff
@@ -61,4 +61,9 @@ public int hashCode() {
     public static Attachment newAttachment(String fileName, String section) {
         return new Attachment(fileName, Date.from(Instant.now()), section);
     }
+    public static Attachment newAttachment(String fileName, String section, String versionId) {
```
millerts: Why do we use a customer builder rather than Lombok's builder?

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -73,20 +87,15 @@
...
         }
         return filenames;
     }
 
-    /**
-     * Returns a list of referenced glossary filenames on the provided Item
-     *
-     * @param item   Item that contains valid attachment references
-     * @return       List of filenames
-     */
```
millerts: Did you mean to remove these comments?

PR #531, opened 2019-11-02
Feature/iat 3429

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -153,13 +153,17 @@ public InputStream getAttachment(String fileKey) {
      * @param unzipFolder The root folder to download the attachments.
      * @return The folder the attachments were unzipped to.
      */
+    //TODO - remove once the zip is removed
```
millerts: You can pretty much ignore this class in the PR. I just went through and deprecated any method that needs to be removed once we've ripped out the archive zip pattern.

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -47,6 +47,30 @@
...
 
+    /**
+     * Given an item a list of all its attachments are returned.  The attachments are enhanced from {@link Attachment}
+     * to {@link DataStoreAttachment}.  The enhancement is to gather enough information so we are able to use the list
+     * to download items from the data store.
+     *
+     * @param item The item.
+     * @return A list of the items attachments.
+     */
+    public static List<DataStoreAttachment> getDataStoreAttachments(Item item) {
```
millerts: I moved this to the top since this is one of the main public methods.  Prior this was mixed with a bunch of package private and private methods making it hard to find.

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -286,14 +286,16 @@
         List<ItemImageResource> imageResources = images.getImageResources();
 
         List<DataStoreAttachment> attachments = imageResources.stream()
-            .filter(imageResource -> Objects.nonNull(imageResource.getProductionFile()))
+            .filter(imageResource -> imageResource.getProductionFile() != null && imageResource.getProductionFile()
```
millerts: This was the bug.  ImageResources have a production and concept attachment object which is not null but has all its properties set to null.  We only care when there is a filename.

PR #532, opened 2019-11-02
Add a flag if the history record supports preview

millerts:


src/main/java/org/opentestsystem/ap/common/model/history/ItemHistory.java
```diff
@@ -44,6 +44,8 @@
 
     private String contentDiff = EMPTY;
 
+    private boolean supportsPreview = false;
```
millerts: When false TIMS will not allow users to preview it.

PR #533, opened 2019-11-04
Feature/iat 3352

bdrainer:
Preview no longer uses the zips.  It uses the recently added DataStoreSaaifContentGenerator.  

There was a lot of clean up done.  I got rid of things that were no longer needed.  There might be more we can get rid of which I'll keep an eye out for and do as part of the next ticket **IAT-3312** _Historical preview should only preview history in TIMS_.

I tested historical preview as part of IAT-3352.  I had to make a few changes to get it to work but it is working.  The ticket I'm about to work on, IAT-3312, requires more changes as I realize we need to consider the new flag in the history table of what is preview-able and what is not.


bdrainer:
I should mention there are breaking changes in this PR.  IRS and IMS both must be updated, PRs coming up.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -184,61 +187,99 @@ public Path downloadAttachmentsZip(String itemId, String versionId, Path downloa
...
+        List<DataStoreAttachment> itemAttachments = AttachmentSelector.getDataStoreAttachments(item);
+
+        // filter out files already downloaded from a non-master branch, we don't want to overwrite it
+        // since the goal is to use the branch attachment file and not the master attachment file.
+        itemAttachments.stream()
+            .filter(attachment -> !branchAttachments.contains(attachment.getAttachmentPath()))
+            .forEach(attachment ->
+            this.downloadAttachment(item.getId(), BRANCH_MASTER, attachment, itemFolder));
+
+        watch.stop(log, format("completed downloading attachments, item %s, branch %s", item.getId(), branch));
```
bdrainer: This is where we are handling branch edits.  We pull the branch files from S3 then pull the master files.  We skip any master files that have the same name as those already pulled from the branch.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -353,7 +372,7 @@ public ItemEntity commitItem(Item branchItem,
                                  ItemBankUser updatedBy) {
         // never commit master on master, that is a patch
         if (!BRANCH_MASTER.equalsIgnoreCase(branch)) {
-            if(StringUtils.isBlank(attachmentCommitResult.getCommitZipVersionId())) {
+            if (StringUtils.isBlank(attachmentCommitResult.getCommitZipVersionId())) {
                 throw new SystemException(
                     "Cannot commit item " + branchItem.getId() + " because the attachment version ID is null");
```
bdrainer: this can likely go away

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -419,6 +439,15 @@ public ItemSyncEntity saveItemSync(ItemSyncEntity itemSync) {
...
 
+    // ------------------------------------------------------------------------
+    // ItemHistory methods
+    // ------------------------------------------------------------------------
+
+    @Transactional(readOnly = true)
+    public ItemHistoryEntity findItemHistoryByUUID(String uuid) {
+        return this.itemHistoryRepository.findOne(UUID.fromString(uuid));
+    }
+
```
bdrainer: When previewing a historic version of an item we are given the item history entity PK.  We look up that entity so we know which item version to use.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -154,12 +149,31 @@ public ItemEntity findCurrentUserEdit(String itemId, String createdBy, String up
         // first try getting branch edits (i.e. non master / item not being created)
         final ItemEntity entity = this.itemEntityRepository.findCurrentUserBranchEdit(itemId, createdBy, updatedBy);
         // if the user is not editing the item, try getting the create record
-        return Objects.nonNull(entity) ? entity : this.itemEntityRepository.findCurrentUserMasterCreate(itemId, createdBy, updatedBy);
+        return Objects.nonNull(entity) ? entity : this.itemEntityRepository
```
millerts: General question, I see this usage of Objects.nonNull().  Is there a reason we use this static method rather than doing `object !=null`?

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -154,12 +149,31 @@ public ItemEntity findCurrentUserEdit(String itemId, String createdBy, String up
...
+    /**
+     * Gets an item by its database PK.
+     *
+     * @param uuid The item's database PK.
+     * @return The item entity
+     */
     @Transactional(readOnly = true)
-    public ItemEntity findItemEntityById(UUID id) {
-        return this.itemEntityRepository.findOne(id);
+    public ItemEntity findItemEntityById(UUID uuid) {
```
millerts: Just my two cents, but I liked the name id regardless of the object type

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -419,6 +439,15 @@ public ItemSyncEntity saveItemSync(ItemSyncEntity itemSync) {
         return this.itemSyncRepository.save(itemSync);
     }
 
+    // ------------------------------------------------------------------------
+    // ItemHistory methods
+    // ------------------------------------------------------------------------
+
+    @Transactional(readOnly = true)
+    public ItemHistoryEntity findItemHistoryByUUID(String uuid) {
```
millerts: Recommend changing this to findItemhistoryById.  Unless the Entity property name is uuid.

src/main/java/org/opentestsystem/ap/common/model/attachment/DataStoreAttachment.java
```diff
@@ -21,6 +22,17 @@
...
     private String folder;
 
+    /**
+     * The data store file key follows the pattern '{itemId}/{branch}/{attachmentPath}'.
+     *
+     * @param itemId The item ID
+     * @param branch The branch
+     * @return The data store file key.
+     */
+    public String getDataStoreFileKey(String itemId, String branch) {
```
millerts: Why does the manager have this public method?  It seems like DataStoreUtility has most of these types of methods.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
bdrainer: You asked this on a different PR, maybe one last week.  I had a longer comment on that PR.  In short,  I prefer Object.nonNull over object != null.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
bdrainer: I like it.  I'll name this findItemById and then somewhere below I will make the history one be findItemHistoryById

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
bdrainer: I will make this change.

src/main/java/org/opentestsystem/ap/common/model/attachment/DataStoreAttachment.java
bdrainer: I'll take a closer look.  It's a good point.

PR #538, opened 2019-11-05
Add a new method with clearer name deprecating the old method

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreSaaifContentGenerator.java
```diff
@@ -44,12 +44,42 @@ public DataStoreSaaifContentGenerator(ItemBankProperties itemBankProperties,
         this.imageGenerator = imageGenerator;
     }
 
+    /**
+     * @deprecated Please use generateSaaifWithAttachments instead
+     */
+    @Deprecated
```
millerts: I don't want to break downstream projects so leaving this here but deprecating.

PR #539, opened 2019-11-05
Removing the zip archive of content from the application.  Also added…

millerts:
… timing logging as the save of larger items can take some time


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -386,20 +383,25 @@ public ItemEntity commitItem(Item branchItem,
...
 
-            jsonMerge.getServerManaged().setTimsAttachmentsVersionId(attachmentCommitResult.getCommitZipVersionId());
-
             dataStoreUtility.updateServerManagedFields(jsonMerge);
             dataStoreUtility.updateRubric(jsonMerge);
+
+            Instant attachmentStart = Instant.now();
             itemAttachmentMapper
                 .updateItemAttachments(jsonMerge, attachmentCommitResult.getMasterAttachmentSummaries());
+            log.info("Attachment update took {} millis", Duration.between(attachmentStart, Instant.now()).toMillis());
```
millerts: I added a bunch of timing logging because performance is poor for items with a lot of attachments.  However, I'm also seeing the item entity being fetched multiple times for an item so there is some inefficiency in the way we query data.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -456,6 +298,7 @@ private String commitAttachments(Item item) {
      * @return the {@link AttachmentCommitResult}
      */
     public AttachmentCommitResult commitItemAttachments(Item item, String branch) {
+        StopWatchLogger watch = StopWatchLogger.start();
```
millerts: I'm a little concerned about the performance of this method as for an item with a lot of terms can take upwards of thirty seconds.  What is surprising is saving the item outside of that can take an additional 20 seconds. I'll have to review tomorrow.

PR #540, opened 2019-11-05
Bugfix/iat 3446

millerts:


src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -56,45 +54,35 @@
...
     private static final boolean YES_EDITING = true;
 
-    private static final boolean FINISH_EDITING = false;
-
     private final static String AccessibilityUpload = "AccessibilityUpload";
 
     private final ItemBankProperties itemBankProperties;
 
-    private final ItemEntityRepository entityMgr;
+    private final ItemEntityRepository itemEntityRepository;
```
millerts: I think we should not abbreviate instance variable names.  Changed this to better represent the object type and make the entire class more readable.

src/main/java/org/opentestsystem/ap/common/bulkupdate/ItemBulkUpdateItemManager.java
```diff
@@ -164,8 +151,7 @@ public void updateBraille(ItemBankUser user,
 
         // begin editing the branch
         ItemEntity entityBranch = this
-            .beginItemEdit(entityMaster, itemId, SECTION_BRAILLE, updateOwner, updateOwnerFullName,
-                "Begin bulk Braille change");
+            .beginItemEdit(entityMaster, itemId, SECTION_BRAILLE, updateOwner, systemUserName());
```
millerts: The issue in this flow that I ran into a few times is that the begin item edit and commit methods were using system user.  I had a mix due to me consolidating the item commit logic to all be in DataStoreItemManager but I consolidating everything hopefully making it clear that system user is being used to save data.  The other method variables were unused so I removed them.

src/test/java/org/opentestsystem/ap/common/security/PermissionValidatorTest.java
```diff
@@ -71,4 +73,38 @@ public void validateCanNotEditGlossaryInAccessibilityUpload() {
         // Forbidden exception will be generated
         PermissionValidator.validateCanEditItemSection(user, item, ModelConstants.Section.SECTION_GLOSSARY);
     }
+
+    @Test
```
millerts: This class needs a lot more tests as if anything goes wrong in the permissions validator it blocks user abilities.

src/main/java/org/opentestsystem/ap/common/security/PermissionValidator.java
```diff
@@ -52,7 +52,7 @@ public static void validateCanPatchItemProperty(ItemBankUser user, Item item, St
     }
 
     public static void validateCanPatchItemMetadataProperty(ItemBankUser user, Item item, String propertyPath) {
-        if (user.can().editItemMetadata(item.getWorkflow().getWorkflowStatusCode())) {
+        if (!user.can().editItemMetadata(item.getWorkflow().getWorkflowStatusCode())) {
```
millerts: The check was inverted.

PR #541, opened 2019-11-05
Update backend permissions for closed captioning edits

burzum619:
Adding additional permissions checks for resolving whether or not a user can edit Closed Captioning content.


src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
```diff
@@ -208,7 +208,10 @@ public boolean editItemCcAudioFileAssociation(String workflowStatusCode) {
     }
 
     public boolean editItemCcContent(String workflowStatusCode) {
-        return this.has("edit_item_cc_content", workflowStatusCode);
+        return this.has("edit_item_cc_content", workflowStatusCode)
```
millerts: There is a test for this class. It is a bit light so nbd

src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
```diff
@@ -208,7 +208,10 @@ public boolean editItemCcAudioFileAssociation(String workflowStatusCode) {
     }
 
     public boolean editItemCcContent(String workflowStatusCode) {
```
bdrainer: I would leave `editIemCcContent` as is and create a new method in Capabilities called `editItemCc` which does exactly what you implemented with the 4 different OR checks.    

`editItemCc` is then applied around the showing the `Edit` button 

`editItemCcContent` is already applied on teh CC page.  If you have this permission then you are able to upload CC files. On  `item-cc-tab.component.html` you'll see around line 26 the `editItemCcContent` check.

burzum619: I like this idea. Thanks!

src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
```diff
@@ -207,6 +207,13 @@ public boolean editItemCcAudioFileAssociation(String workflowStatusCode) {
         return this.has("edit_item_cc_audio_file_association", workflowStatusCode);
     }
 
+    public boolean editItemCc(String workflowStatusCode) {
+        return editItemCcContent(workflowStatusCode)
+            || editItemCcIsProvided(workflowStatusCode)
+            || editItemCcIsRequired(workflowStatusCode)
+            || editItemCcAudioFileAssociation(workflowStatusCode);
+    }
+
```
bdrainer: perfect

PR #542, opened 2019-11-05
IAT-3268: TIMS UI - the history tab does not display the 'Began creat…

bdrainer:
…ion' entry.  I first used the `lockItem` method but that required putting a transaction around the code to read the history.  It just seems better not having a transaction that locks the item when all we are doing is reading data.


src/main/java/org/opentestsystem/ap/common/datastore/repository/ItemEntityRepository.java
```diff
@@ -32,6 +32,21 @@
...
+     * @param itemId The unique item identifier.
+     * @return The first record inserted (i.e. the oldest record) for an item.
+     */
+    @Query("SELECT entity FROM ItemEntity entity WHERE entity.itemId = :itemId"
+        + " AND entity.branchName = '" + BRANCH_MASTER + "'"
+        + " AND entity.createdDate = "
+        + "     (SELECT MIN(d.createdDate) FROM ItemEntity d WHERE d.itemId = :itemId "
+        + "         AND d.branchName = '" + BRANCH_MASTER + "')"
+    )
+    ItemEntity findFirstMaster(@Param("itemId") String itemId);
```
millerts: This works but could you do the following:

```
select entity FROM ItemEntity entity WHERE entity.branchName = 'master' ORDER BY entity.createdDate ASC limit 1;
```

Not even sure you need the branch name equals master since every item in tims is created on master.

bdrainer: I like it.  I tried the Spring query with LIMIT but no dice.  I just discovered 
`@Query(value="SELECT * FROM ITEM WHERE ITEM_ID = ?1 ORDER BY CREATED_DATE ASC LIMIT 1", nativeQuery = true)` .  Native query, who knew.

PR #543, opened 2019-11-06
Add logic to handle logic around tutorial attachments

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -290,6 +290,19 @@ public void downloadAttachment(String fileName, Item item, String branch, Path r
...
 
+    /**
+     * Finds the {@link AttachmentSummary} for the item based on branch.  The collection only contains the attachments
+     * for the given branch.
+     *
+     * @param item   {@link Item} with attachments
+     * @param branch the branch to look at for attachments
+     * @return {@link List} the attachment summaries for the branch
+     */
+    public List<AttachmentSummary> findAttachmentSummaries(Item item, String branch) {
```
millerts: This is used by the migration

PR #544, opened 2019-11-06
Fix the attachment selector so it doesn't create null attachments

millerts:
Added guard code to throw better excptions when the filename is null


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -470,6 +470,10 @@ public void downloadItemFiles(List<String> fileNames, String itemId, String bran
      * @param rootPath Path where files are copied to
      */
     private void downloadItemFile(String fileName, String itemId, String branch, Path rootPath) {
+        if(StringUtils.isEmpty(fileName)) {
```
millerts: I added guards here because under the hood the exception that is thrown is a vague S3 exception that it can't find hte object, but it doesn't give us the object name.

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -69,7 +69,10 @@
             attachments.addAll(getAttachments(item.getSaaif()));
             attachments.addAll(getItemTypeSpecificAttachments(item));
         }
-        return attachments;
+
+        return attachments.stream()
```
millerts: Somewhere in this selector it doesn't handle the item structures being created.  Instead of trying to find the permutations everywhere i just clean up the list before sending it out.  I no longer see the Exceptions in the other services I saw failing previously.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
bdrainer: Nice, this is going to be helpful

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
bdrainer: makes sense

PR #545, opened 2019-11-06
IAT-3443 - Bulk assign function does not respect 'assign' permissions

alex-ponce:
Add validation to prevent attempting to bulk assign/unassign items that are being created


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -342,6 +343,9 @@ public ItemResponse patchItem(ItemBankUser user, ItemPatchRequest request) {
         this.dataManager.lockMaster(itemId);
 
         ItemEntity originatingEntity = this.dataManager.findLastMaster(itemId);
+        if (originatingEntity.isBeingCreated()) {
+            throw new SystemException(String.format("Item %s is being created. Items in this state cannot be bulk assigned", itemId));
```
millerts: Does this message get relayed to the screen either in a dialog or growl?

alex-ponce: Yes it does.

PR #546, opened 2019-11-06
Change the code to only save the version id on new attachments

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -315,40 +316,50 @@ public AttachmentCommitResult commitItemAttachments(Item item, String branch) {
...
 
-                return !masterAttachment.getMd5Hash().equals(branchAttachmentSummary.getMd5Hash());
-            }).map(AttachmentSummary::getKey)
-                .collect(Collectors.toList());
+            if(!branchObjectToCopy.isEmpty()) {
+                // sync the branch folder to the master folder, non-versioned bucket
+                Instant startBranchCopyFolder = Instant.now();
+                this.copyBranchFolderToMasterFolder(itemId, branch,
+                    branchObjectToCopy.stream().map(AttachmentSummary::getKey).collect(
+                        Collectors.toList()));
```
bdrainer: I wonder if there is a way that `this.copyBranchFolderToMaster` can return a list of AttachmentSummary?  We are calling `client.copyObject` that returns aCopyObjectResult.  That has the version ID and eTab.  Doing this would avoid needing to call `s3BucketManager.findAttachmentSummary` below around line 344.

PR #547, opened 2019-11-06
Fix the selector as the filter was inverted.

millerts:


src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -71,8 +71,8 @@
         }
 
         return attachments.stream()
-            .filter(dataStoreAttachment -> StringUtils.isEmpty(dataStoreAttachment.getAttachmentPath()) || StringUtils
-                .isEmpty(dataStoreAttachment.getAttachment().getFileName())).collect(Collectors.toList());
+            .filter(dataStoreAttachment -> StringUtils.isNotEmpty(dataStoreAttachment.getAttachment().getFileName()))
```
millerts: Running with scissors and I got cut.  The conditional was inverted.  We only want attachments that have filename.  I also realized that getAttachmentPath wasn't the correct thing to use since does some string concat

PR #548, opened 2019-11-09
Refactor Migrations to use S3 rather than Gitlab

millerts:
During staging deployment Brett mentioned Migrations were still pulling hte import.zip from Gitlab.  This removes that Gitlab dependency


src/main/java/org/opentestsystem/ap/common/saaif/mapper/GiModelMapper.java
```diff
@@ -37,6 +38,10 @@
 @Slf4j
 @NoArgsConstructor
 public class GiModelMapper extends IatModelMapper {
+    //Both these static methods are used in older Migrations in addition to their usage here
+    public static final String GI_SAMPLE_DIRECTORY_NAME = "gi-sample";
+    public static final String GI_SAMPLE_FILE_NAME = "gi-sample.zip";
+
```
millerts: Moved these from MigrationFileUtil because they're only needed for GI items

src/main/java/org/opentestsystem/ap/common/saaif/mapper/util/MigrationFileUtil.java
```diff
@@ -182,6 +121,25 @@ public Path getExpandedImportZipFile(String itemId, byte[] bytes) {
                 itemFile));
     }
 
+    public Optional<Path> getExpandedImportZipFile(final ItemEntity itemEntity) {
```
millerts: This is a refactoring of the existing getExpandedImportZip file which fetches the item from S3.

src/main/java/org/opentestsystem/ap/common/saaif/mapper/util/MigrationFileUtil.java
```diff
@@ -45,74 +43,15 @@
...
+    private final DataStoreAttachmentManager dataStoreAttachmentManager;
 
-    public MigrationFileUtil(final GitlabClient gitlabClient,
-                             final AppAssembler appAssembler,
-                             final ItemRepository itemRepository) {
-        this.gitlabClient = gitlabClient;
+    public MigrationFileUtil(final AppAssembler appAssembler,
+                             final DataStoreAttachmentManager dataStoreAttachmentManager) {
         this.appAssembler = appAssembler;
-        this.itemRepository = itemRepository;
```
millerts: Red in this class is all unused code.

src/main/java/org/opentestsystem/ap/common/saaif/mapper/GiModelMapper.java
bdrainer: Makes sense

PR #551, opened 2019-11-10
Feature/fix history importer

millerts:


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreDataManager.java
```diff
@@ -441,6 +441,16 @@ public ItemSyncEntity saveItemSync(ItemSyncEntity itemSync) {
         return this.itemSyncRepository.save(itemSync);
     }
 
+    @Transactional
+    public ItemHistoryEntity saveItemHistory(ItemHistoryEntity itemHistory) {
```
millerts: I can move this back to the history section.  I didn't realize the comment blocks were breaking the chunks of DB access.

src/main/java/org/opentestsystem/ap/common/util/StopWatchLogger.java
```diff
@@ -40,6 +40,6 @@ public void reset() {
     }
 
     private String formatMessage(String message) {
-        return String.format("TIMS_PERFOMRANCE %s - %d", message, this.watch.getTime());
+        return String.format("TIMS_PERFOMRANCE %s - took %d millis", message, this.watch.getTime());
```
millerts: It wasn't clear to me without looking at the underlying implementation what the number represented in time units.  Added a bit of extra logging to help

PR #554, opened 2019-11-13
Change it so that if an item cannot be found in s3 a null is returned

millerts:
rather than throwing an error


src/main/java/org/opentestsystem/ap/common/config/DatabaseConfiguration.java
```diff
@@ -120,14 +120,9 @@ public DataStoreDataManager dataStoreDataManager(ItemBankProperties itemBankProp
 
     @Bean
     @ConditionalOnMissingBean
-    public DataStoreAttachmentManager dataStoreAttachmentManager(ItemBankProperties itemBankProperties) {
+    public DataStoreAttachmentManager dataStoreAttachmentManager(final ItemBankProperties itemBankProperties) {
         log.info("Creating bean dataStoreAttachmentManager");
-        return new DataStoreAttachmentManager(
```
millerts: I removed a bunch of unused instance variables so cleaned up this constructor.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -45,27 +41,13 @@
 
     private static final String NULL_VERSION_ID = null;
 
-    private static final String ITEM_ATTACHMENTS_ZIP_PATTERN = "%s-attachments.zip";
```
millerts: All red is unused.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -84,6 +66,8 @@ public void saveAttachment(String fileKey, Path filePath) {
         s3BucketManager.uploadFile(fileKey, filePath);
     }
 
+    //TODO - this should be removed as it is leaking S3 internals outside of the data layer
+    @Deprecated
```
millerts: I didn't try to find where this is being used due to the timeframe of the class, but I don't think this method should be here or public.  Ideally, the fact TIMS uses S3 should be abstracted away so the system doesn't care if we're use S3, Azure, local file system

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
```diff
@@ -79,23 +78,6 @@ public void initialize() {
             .build();
     }
 
-    public String zipAndSave(Path dataToZip, Path zipFile) {
```
millerts: Unused

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
```diff
@@ -115,13 +97,24 @@ public String uploadFile(String fileKey, Path filePath) {
         }
     }
 
+    /**
+     * Downloads the file from S3 with the file key and version id
+     *
+     * @param fileKey   file key for the file (required)
+     * @param versionId version id for the file.  If not provided will return the latest
+     * @return {@link InputStream} if found otherwise null
+     */
```
millerts: This method is where the main change happened.  After talking with the client we feel this is the safest course of action for the time being.  Smarter found other imported items that have references in the item.json which are not in S3.  The latest one even has the file within the import.zip.

At a future date Smarter will revisit if they'd like us to audit the data in the system, but for now the tabulator checks for invalid file references allowing users to clean up the import issues.

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
```diff
@@ -148,7 +142,8 @@ public InputStream downloadFile(String fileKey) {
      * @param versionId The version of the file to download.
      * @return The S3 resource information including the file input stream.
      */
-    protected Optional<S3Object> downloadFromS3(String fileKey, String versionId) {
+    @VisibleForTesting
+    Optional<S3Object> downloadFromS3(String fileKey, String versionId) {
```
millerts: This only seemed to be used in testing so reduced scope.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
bdrainer: This is used by IMS when getting the resource from the QTI call.  There is a ticket I'm working here shortly to handle getting historic versions of resources.  I'll fix IMS so it gets a summary object instead of a S3 specific object.

PR #555, opened 2019-11-13
Include commitId when returning ItemTransactions for in-flight edits

wtritch:
The QTI item previewer needs the commit ids for in-flight changes in order to request the appropriate item json and binary resources.

This PR adds the associated commitId when returning "transactions" for an item.

image

Note that this property may be NULL if the ItemTransaction instance is provided via the GitLab client.


src/main/java/org/opentestsystem/ap/common/client/GitlabClient.java
```diff
@@ -144,7 +144,10 @@ public ItemTransaction lookupTransactionByUsername(final Project project, final
                     .map(Branch::getName)
                     .filter(name -> !equalsIgnoreCase(name, BRANCH_MASTER))
                     .map(name -> split(name, UNDERSCORE, 2))
-                    .map(split -> new ItemTransaction(split[1], split[0]))
+                    .map(split -> ItemTransaction.builder()
+                        .userName(split[1])
+                        .section(split[0])
+                        .build())
```
wtritch: So if the ItemTransaction is coming from the Gitlab client, we don't have access to a commitId value.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -97,7 +97,7 @@ public ItemResponse createItem(ItemBankUser user,
                                    String workflowStatus) {
         ItemEntity masterEntity = this.dataManager.createItem(user, itemId, itemType, isBeingEdited, workflowStatus);
         this.eventProducer.sendCreateEvent(user, masterEntity, commitMessage);
-        return new ItemResponse(masterEntity.getItemJson(), newCreateTransaction(user.getUserName()));
+        return new ItemResponse(masterEntity.getItemJson(), newCreateTransaction(masterEntity));
```
esotrope: new create transaction x J

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
```diff
@@ -250,7 +254,7 @@ public ItemResponse findItem(ItemBankUser user, String itemId) {
         if (Objects.isNull(masterEntity)) {
             throw new ResourceNotFoundException("Item " + itemId + " not found");
         } else if (this.dataStoreUtility.isUserCreatingItem(user, masterEntity)) {
-            itemResponse = new ItemResponse(masterEntity.getItemJson(), newCreateTransaction(user.getUserName()));
+            itemResponse = new ItemResponse(masterEntity.getItemJson(), newCreateTransaction(masterEntity));
```
wtritch: It looks like we only call `newCreateTransaction` when actively creating the item, here and above on line 100.  The ItemEntity instance has all the details needed to create the initial "creation" ItemTransaction.

src/main/java/org/opentestsystem/ap/common/client/GitlabClient.java
millerts: Hopefully this is never used by QTI.  GItlab will be going away in the next few months.  I realize this is probably a refactor type of edit.

src/main/java/org/opentestsystem/ap/common/model/ItemTransaction.java
```diff
@@ -1,22 +1,39 @@
...
+import javax.annotation.Nullable;
+import java.util.UUID;
 
 import static org.opentestsystem.ap.common.model.ModelConstants.BranchNames.BRANCH_MASTER;
 import static org.opentestsystem.ap.common.model.ModelConstants.Section.SECTION_CREATE;
 
-@Data
+@Getter
+@Builder
 public class ItemTransaction {
```
wtritch: With the third property, I switched this from an all-args constructor to a builder with getter methods.

src/main/java/org/opentestsystem/ap/common/datastore/DataStoreItemManager.java
wtritch: Yeah, removed the helper method in favor of using the builder

src/main/java/org/opentestsystem/ap/common/client/GitlabClient.java
bdrainer: I want to say this is old and we never get a transactions from GitLab.  We used to, but not now since moving to the data store.

PR #556, opened 2019-11-14
IAT-3307 - TIMS translation service content package endpoint does not produce identical output to the legacy tool

alex-ponce:
Added logic to identify illustration images if they are referenced on a glossary term.


src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -127,6 +127,16 @@
...
+            // Add illustration image if referenced
+            if (StringUtils.isNotBlank(term.getIllustrationImageResourceId())) {
+                filenames.addAll(item.getImages().getImageResources().stream()
+                    .filter(itemImageResource ->
+                        itemImageResource.getId().equals(term.getIllustrationImageResourceId())
+                            && Objects.nonNull(itemImageResource.getProductionFile()))
+                    .map(itemImageResource -> itemImageResource.getProductionFile().getFileName())
+                    .collect(Collectors.toList()));
+            }
+            // Add language dialect audio files if referenced
```
alex-ponce: Return image file name if it is referenced as a glossary illustration

millerts: After @bdrainer 's work we now have two methods that find attachments, correct?  Can expand on the javadoc you have to provide more detail?  It does fetch attachments on the item, but the method Brett has does as well. This flow seems to go to the extra step of only pulling what is referenced in content?

If there really is no difference can we refactor the methods to have a single on?

PR #557, opened 2019-11-18
Feature/iat 3475

bdrainer:
This is a significant PR.  It removes dependencies from common which affects all services dependent on common.  The merging of this will need coordinated as all other services will need adjusted.

The dependencies removed are:  spring data redis, spring session, and spring security saml.  There are two exceptions (IAT and API Gateway), but services are now stateless and no longer require spring session/spring data redis.  These services used to be configured with basic security however, now they use a JWT web security configuration that puts a filter in front of requests for decoding a JWT authentication header.  The header is decoded and put into the security context.  Once the security context is set the services behave as they always have where they use SecurityUtil to get the user along with their permissions.

The new code was copied from RDW.  I tweaked the ZuulFilter to put the token for all requests which differs from RDW.  RDW has a pattern matcher that checks if it should put the token in the header or not.  I didn't think TIMS needed that but maybe it does given the viewer service.  Maybe there are some things the TIMS UI Zuul proxy routes to the viewer service that does not require the token.  I think we are ok with putting it on every request but I wanted to mention it in case others can think of reasons we should have it.


src/main/java/org/opentestsystem/ap/common/security/authentication/PermissionServiceClient.java
```diff
@@ -17,10 +20,11 @@
     public PermissionServiceClient(RestTemplateBuilder restTemplateBuilder,
                                    ItemBankProperties.PermissionServiceProps properties) {
         this.appProperties = properties;
-        this.restTemplate =  restTemplateBuilder.rootUri(properties.getUrl()).build();
+        this.restTemplate = restTemplateBuilder.rootUri(properties.getUrl()).build();
     }
 
     public AuthenticatedUserModel findAuthenticatedUser(String username) {
+        log.info("find authenticated user {}", username);
```
millerts: Can we change this to debug?

src/main/java/org/opentestsystem/ap/common/security/jwt/DefaultJwtService.java
```diff
@@ -0,0 +1,64 @@
...
+                .getBody()
+                .get(UserClaim)
+        );
+    }
+
+    private String userToJson(final User user) {
+        try {
+            return objectMapper.writeValueAsString(user);
+        } catch (final JsonProcessingException exception) {
+            throw new RuntimeException(exception);
```
millerts: Should this throw SystemException since that is the TIMS runtime exception?

src/main/java/org/opentestsystem/ap/common/security/jwt/JwtSecurityConfiguration.java
```diff
@@ -0,0 +1,36 @@
...
+
+    @Bean
+    @ConfigurationProperties(prefix = "jwt")
+    public JwtSettings jwtSettings() {
+        return new JwtSettings();
+    }
+
+    @Bean
+    public JwtService jwtService(final JwtSettings settings) {
+        final ObjectMapper mapper = new ModelObjectMapper();
```
millerts: Should TIMS register a single mapper that all classes use?  Is there any harm in wiring up a mapper that has all these properties and modules that all systems can use or is there a side effect?

src/main/java/org/opentestsystem/ap/common/security/model/UserBuilder.java
```diff
@@ -35,7 +37,7 @@
 
     public static UserBuilder builder() { return  new UserBuilder(); }
 
-    private UserBuilder() {};
+//    private UserBuilder() {};
```
millerts: delete?

src/main/java/org/opentestsystem/ap/common/config/ObjectMapperConfiguration.java
```diff
@@ -1,23 +1,17 @@
...
 
     @Bean
     public Jackson2ObjectMapperBuilderCustomizer customizeObjectMapper() {
         return builder -> builder
             .failOnUnknownProperties(false)
-            .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
-            .featuresToEnable(MapperFeature.SORT_PROPERTIES_ALPHABETICALLY)
-            .modulesToInstall(new Jdk8Module(), new JavaTimeModule());
-
+            .featuresToDisable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS);
```
millerts: Not sure how this is used elsewhere but don't we need the modules installed?

src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
```diff
@@ -645,7 +645,7 @@ public boolean viewWerItemAssignedToYou(String workflowStatusCode) {
         return this.has(viewWerItemAssignedToYou, workflowStatusCode);
     }
 
-    private boolean has(String permissionCode, String workflowStatusCode) {
+    boolean has(String permissionCode, String workflowStatusCode) {
```
millerts: If this was changed from private to package private I assume it is because of a test?  If so, can you add the VisibleForTesting annotation?   Or maybe it should be protected since it seems you added this for inheritance?

src/main/java/org/opentestsystem/ap/common/config/ObjectMapperConfiguration.java
bdrainer: I'm not sure and I struggled with this configuration.  I took it from ISS or IIS, you'll see it deleted there.  I had the modules registered, though I had more than just those, I also had the ones from `JwtSecurityConfiguration` where it configures an object mapper for the the `JwtService`.  Everything seemed to be fine but tests started to fail and applications started to break.  I'm not clear what happened but one of those modules caused the JSON being generated to look odd. 

I might have had too much going on but it took me longer than should have been necessary to get things back working.  You might recall the CI Slack channel and the continuous test failures.  That was me battling this configuration.

With all that being said I can put those two modules back (Jdk8 and JavaTime).   It seems unlikely those two are the problem but I feel like I did test that.  Ultimately I threw up my hands and went with what I could get working.

src/main/java/org/opentestsystem/ap/common/model/admin/Capabilities.java
bdrainer: I will make this protected.  It was for inheritance.

src/main/java/org/opentestsystem/ap/common/config/ObjectMapperConfiguration.java
bdrainer: I added back the two modules and things are working as expected.

src/main/java/org/opentestsystem/ap/common/model/admin/SystemCapabilities.java
```diff
@@ -0,0 +1,17 @@
...
+        super(Collections.emptyList());
+    }
+
+    /**
+     * System capabilities have no restrictions.
+     */
+    protected boolean has(String permissionCode, String workflowStatusCode) {
+        return true;
+    }
+}
```
bdrainer: The idea here is services like IIS where they need to call other services can utilize the `ItemBankSystemUser` which uses `SystemAuthenticatedUserModel` that internally uses `SystemCapabilities`.  When IIS makes a call it will encrypt   `SystemAuthenticatedUserModel` and put it as the authorization header.  Receiving services like IMS will decrypt the user where it will be the `SystemAuthenticatedUserModel`.

The main thing to notice is `SystemAuthenticatedUserModel` **has all capabilities**.  It is a super user.  Is this what we want?  It seems like it is.   A different approach would be to add system users to the database.  Services could load their system user using the permissions service.

src/main/java/org/opentestsystem/ap/common/restclient/AppendAuthenticationHeaderInterceptor.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.io.IOException;
+
+/**
+ * Uses this interceptor when there is a user in the security context.  The interceptor is applied to a {@link
+ * org.springframework.web.client.RestTemplate} where it will encode the user in the security context and set it on the
+ * outbound request as the header property "Authorization".
+ */
+@Slf4j
+public class AppendAuthenticationHeaderInterceptor implements ClientHttpRequestInterceptor {
```
bdrainer: This interceptor is used by services that pass through authenticated user requests.  For example, the api proxy forward to the translation service and the translation service forwards to IMS.  All three services will have an authenticated user in their security context.  This interceptor utilizes the security context to retrieve the user, encrypt it, and add it as the Authorization header to the outbound request.

src/main/java/org/opentestsystem/ap/common/restclient/AppendSystemAuthenticationHeaderInterceptor.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.http.client.ClientHttpResponse;
+
+import java.io.IOException;
+
+/**
+ * Uses this interceptor when a system needs to make authenticated calls into another service.  The interceptor is
+ * applied to a {@link org.springframework.web.client.RestTemplate} where it will encode the system user.
+ */
+@Slf4j
+public class AppendSystemAuthenticationHeaderInterceptor implements ClientHttpRequestInterceptor {
```
bdrainer: This interceptor is used when a service needs to call a different services as a system user.  This happens when the calling system is running outside the context of a user request.  IIS is a prime example.  It is a schedule job.

PR #558, opened 2019-11-19
Expand GI archive on upload

wtritch:
Record GAX and referenced GI resources in item.json

This PR just expands the GI Json model to embed the GAX interaction definition xml and include references to uploaded GI resources (background images, draggable images, spanish-translation-override images, etc)

GI Archive expansion PR set:
AP_ItemManagementService: https://github.com/SmarterApp/AP_ItemManagementService/pull/189
AP_Common changes: https://github.com/SmarterApp/AP_Common/pull/558
AP_ItemAuthoringTool changes: https://github.com/SmarterApp/AP_ItemAuthoringTool/pull/786


src/main/java/org/opentestsystem/ap/common/datastore/attachment/mapper/GiItemAttachmentMapper.java
```diff
@@ -11,12 +11,15 @@
...
 public class GiItemAttachmentMapper extends GeneralItemAttachmentMapper implements ItemTypeAttachmentMapper {
+
     @Override
     public Item updateItemAttachments(final Item item, final Map<String, AttachmentSummary> attachmentSummaries) {
         Item attachmentMappedItem = super.updateItemAttachments(item, attachmentSummaries);
 
         GiItem giItem = (GiItem) attachmentMappedItem;
         updateAttachmentVersionId(giItem.getCore().getEn().getContentZipFile(), attachmentSummaries);
+        giItem.getCore().getEn().getGiResources()
+            .forEach(giResource -> updateAttachmentVersionId(giResource, attachmentSummaries));
```
wtritch: I updated the version-toucher to also update versions for any referenced GI resources uploaded in the ZIP archive.

src/main/java/org/opentestsystem/ap/common/model/content/GiItemContent.java
```diff
@@ -46,12 +72,20 @@ public boolean equals(Object o) {
...
         } else {
             throw new SystemException("Object is not of expected type 'GiItemContent'");
         }
     }
 
+    @Override
+    public int hashCode() {
+        return Objects.hash(getPrompt(), getContentZipFile(), isRequired(),
+            isProvided(), getGiDefinition(), getGiResources());
+    }
```
wtritch: nit: if we override "equals" we *must* override "hashCode" =)

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -186,7 +187,8 @@
             }
 
         } catch (Exception ex) {
-            log.error("Unable to retrieve filenames from gax file", ex);
+            log.warn("Problem parsing filenames from gax file", ex);
+            throw new ValidationException("Problem parsing filenames from gax file", ex);
         }
```
wtritch: If we're attempting to parse an invalid `.gax` file, this should advertise that the XML is invalid.

src/main/java/org/opentestsystem/ap/common/model/content/GiItemContent.java
esotrope: but if you do that it won't work everytime 60% of the time anymore x p

src/main/java/org/opentestsystem/ap/common/datastore/attachment/mapper/GiItemAttachmentMapper.java
```diff
@@ -11,12 +11,15 @@
...
  */
 public class GiItemAttachmentMapper extends GeneralItemAttachmentMapper implements ItemTypeAttachmentMapper {
+
     @Override
     public Item updateItemAttachments(final Item item, final Map<String, AttachmentSummary> attachmentSummaries) {
         Item attachmentMappedItem = super.updateItemAttachments(item, attachmentSummaries);
 
         GiItem giItem = (GiItem) attachmentMappedItem;
         updateAttachmentVersionId(giItem.getCore().getEn().getContentZipFile(), attachmentSummaries);
+        giItem.getCore().getEn().getGiResources()
```
millerts: @bdrainer , the saaif generation code ignores these new resources, correct?  As in this change does not impact saaif preview which right now requires everything on disk and we pull from the sample zip.

bdrainer: That is correct.  The saaif generation will ignore these resources and specifically the AttachmentSelector is not looking at these new references.  I agree this should not be a problem for preview as it use the gi-sample.zip.

PR #561, opened 2019-11-21
Feature/qti 168 version item json on deserialization

wtritch:
This PR adds item JSON versioning with support for successive version-to-version conversions when de-serializing historical item JSON.

The goal is that the item.json can become more flexible, allowing changes to the JSON model moving forwards while allowing downstream systems to always assume they will be provided item JSON as the latest version.

The first version bump (LegacyToVersion45) is basically just a no-op tagging implementation, due to S3 resource changes in GI items between version 44 and 45.


src/main/java/org/opentestsystem/ap/common/model/JsonModelAssembler.java
```diff
@@ -72,7 +72,7 @@ public Item readItem(final Path file) {
 
     public Item readItem(final InputStream inputStream) {
         try {
-            return mapper.readValue(inputStream, AbstractItem.class);
+            return mapper.readValue(inputStream, Item.class);
```
wtritch: I hope there isn't a hard requirement hidden somewhere to be able to deserialize as an AbstractItem... ?  We've only told Jackson how to deserialize an Item.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemDeserializer.java
```diff
@@ -0,0 +1,141 @@
...
+        ItemVersion version = getVersion(node);
+
+        //Sequentially convert the json
+        for (final ItemVersionConverter converter : converters) {
+            if (version.compareTo(converter.getTargetVersion()) < 0) {
+                node = converter.convert(node);
+                version = converter.getTargetVersion();
+                setVersion(node, version);
+            }
+        }
```
wtritch: Here's the magick.  We parse the JSON string as a basic ObjectNode, pass that sequentially to all converters that produce a later version, then de-serialize naturally as the specified Item concrete class.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemDeserializer.java
```diff
@@ -0,0 +1,141 @@
...
+        if (node.has("serverManaged")) {
+            final ObjectNode serverManaged = (ObjectNode) getChild(node, "serverManaged");
+            if (serverManaged.has("version")) {
+                return ItemVersion.parse(getChild(serverManaged, "version").textValue());
+            }
+        }
+        return ItemVersion.parse(getChild(node, "version").textValue());
+    }
+
+    private void setVersion(final ObjectNode node, final ItemVersion version) {
```
wtritch: @millerts and I have discussed that the `$.version` (and/or `$.serverManaged.version`) will now represent the internal JSON schema version, rather than the migration version.  Migration versions will be tracked outside of the item.json inside a new item-to-migration-version lookup table in the DB.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersion.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * This model represents a parsed item JSON version.
+ *
+ * @see Item#getVersion()
+ */
+@Getter
+@EqualsAndHashCode
+@AllArgsConstructor
+@Slf4j
+public class ItemVersion implements Comparable<ItemVersion> {
```
wtritch: Just a simple version class for being able to compare `iat-32.4` to `iat-43` to `gobbldygook` to `null`

esotrope: no way to use a semantic versioning lib for this?

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersionConverter.java
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Implementations of this interface are responsible for converting
+ * Item JSON from a previous version to a new version.
+ *
+ * Implementations are responsible for
+ * generating/removing/moving/splitting/renaming/etc
+ * any properties that have been modified from the prior version.
+ */
+public interface ItemVersionConverter {
```
wtritch: Any change to the item.json structure should have an associated converter responsible for making old json look like new json.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersioningModule.java
```diff
@@ -0,0 +1,32 @@
...
+    public ItemVersioningModule() {
+        super("TIMS Item Versioning Module");
+
+        final ItemDeserializer itemDeserializer = itemDeserializer();
+        super.addDeserializer(Item.class, itemDeserializer);
+    }
+
+    //TODO: Change this to DI if/when we no longer have
+    // static ObjectMapper creation/references.
+    private ItemDeserializer itemDeserializer() {
```
wtritch: This is unfortunate.  We can't use natural Spring DI in our converters mostly due to Hibernate referencing its ObjectMapper via className.
If a future converter needs access to S3 or any other injected service/configuration, it'll need to perform some nasty workarounds like static property injection.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersion.java
wtritch: I didn't see an easy one, but it's not a huge class.
The problem is that we aren't technically using semantic versioning, sometimes there's a minor, sometimes not.  There's never a micro.  But there's always (?) a prefix of `iat-`

millerts: The prefix should go away moving forward.

millerts: And we could enforce semantic versioning moving forward as well if we wanted

agorina: I am curious how the version is being planned to be used (besides differentiating between different models) and what is the meeting of minor/major? Are we overcomplicating it? 

Maybe not, just wanted to share the thought.

millerts: Good point.  The `item.version` property has a long history and is misnamed.  It should have been named migratedVersion which tracks the migration version.  This was added to the item.json due to TIMS initial reliance on Gitlab as the only datastore.  

As I think Alla is hinting at I see no reason why moving forward this version will need to have minor on it.  I would think that the `iat-` will be removed and it will be whole number increments for each model change.  Or at least that is my vote.

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersioningModule.java
millerts: :(

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersion.java
wtritch: done, switched to a flat int version: "45"

src/main/java/org/opentestsystem/ap/common/model/versioning/ItemVersioningModule.java
wtritch: Looks like this is due to our library choice: `com.vladmihalcea:hibernate-types-5:2.2.0`  The library doesn't really provide a large amount of functionality.  If/when a conversion needs injection, it might be worth looking into re-implementing their hibernate json-bindings for json string DB values so that we can inject the one-true ObjectMapper.

PR #563, opened 2019-11-22
Adding some support for expanding GI archives during migration.

wtritch:
Hopefully this is a step towards standardizing some of the attachment handling methods.


src/main/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManager.java
```diff
@@ -457,6 +459,46 @@ public void downloadItemFiles(List<String> fileNames, String itemId, String bran
...
+     * @param attachment    The attachment reference
+     * @param contentLength The content length in bytes
+     * @param inputStream   The attachment contents
+     * @return  The uploaded attachment
+     */
+    public DataStoreAttachment saveAttachment(final String itemId,
+                                              final String branch,
+                                              final DataStoreAttachment attachment,
+                                              final long contentLength,
+                                              final InputStream inputStream) {
```
wtritch: It looked like everything available was based off of Path or MultiPartFile, which leads to some duplication.  I just wanted some simple, re-usable methods based on streams + metadata for uploading/downloading attachments with version support.

src/main/java/org/opentestsystem/ap/common/datastore/client/S3BucketManager.java
```diff
@@ -97,6 +101,46 @@ public String uploadFile(String fileKey, Path filePath) {
...
+
+            final ObjectMetadata metadata = s3Object.get().getObjectMetadata();
+            return new StreamBackedResource(s3Object.get().getObjectContent(), metadata.getContentLength(), metadata.getContentType());
+
+        } catch (final AmazonServiceException e) {
+            if (e.getStatusCode() == 404 && "NoSuchKey".equalsIgnoreCase(e.getErrorCode())) {
+                throw new NoSuchElementException("No resource found for fileKey: [" + fileKey + "] version [" + versionId + "]");
+            }
+            throw new SystemException("Problem fetching binary resource", e);
+        }
```
wtritch: These try/catches are keeping the S3 knowledge down inside the S3BucketManager

src/test/java/org/opentestsystem/ap/common/datastore/DataStoreAttachmentManagerTest.java
```diff
@@ -55,7 +123,6 @@ public void downloadReferencedAttachmentsForSaItem() throws IOException {
         }
     }
 
-    @Ignore
```
esotrope: we know why these were ignored? don't work on CI for instance?

wtritch: Yeah, they hit and modify S3 and require AWS credentials.  Both are no-no's when performing CI builds.

wtritch: (Rather than \@Ignore every test, I moved the annotation to the class)

millerts: There are a mix of tests within TIMS common.  90% of the tests are Integration or Unit Tests that can be run during normal CI builds.  There are about 10% of tests which I wouldn't necessarily call tests in the traditional sense but more like full End to End tests like what Will is referencing that actually goes and creates records in S3, Gitlab, etc.

We probably need to pull those tests into their own class types and leave the unit test naming convention to mean unit test.  Not asking you to do this @wtritch ; just mentioning where we need to get to long term.

PR #564, opened 2019-12-05
Fix custom operator names to align with TDS_ItemScoring

wtritch:
While mapping TIMS JSON operator representations to QTI operators I noticed that the names of these two operators don't match what is defined in TDS_ItemScoring.

Associated mapping PR (for reference:) https://github.com/SmarterApp/AP_ItemTranslationService/pull/45


src/main/java/org/opentestsystem/ap/common/model/ModelConstants.java
```diff
@@ -185,8 +185,8 @@
         String GetValueNumeric = "GETVALUENUMERIC";
         String GetVector = "GETVECTOR";
         String HasVertex = "HASVERTEX";
-        String IntersectPoint = "INTERSECTPOINT";
-        String IntersectRegion = "INTERSECTREGION";
+        String IntersectPoint = "INTERSECTSPOINT";
+        String IntersectRegion = "INTERSECTSREGION";
```
wtritch: In general, I believe we want to match TDS_ItemScoring in here.  (Which is why "ExpressionContains" is implemented with the TDS_ItemScoring misspelled "EXRESSIONCONTAINS"

wtritch:
There are also some minor typing changes we may want to look into in the future:
```
CustomStringToFloatExpression
 - spec calls out a "inputString" parameter: A string formatted as a numerical value

CustomGiGetPointExpression
 - pointIndex attribute should be an Integer rather than a String (minor nit, seems to not be used by the scoring engine)

CustomGiIntersectPointExpression
 - x and y should be Integer
 - tolerance should be Double

CustomGiIntersectRegionExpression
 - left-x, top-y, etc should be Integer rather than String

CustomGiGetVectorExpression
 - order should be Integer rather than String

CustomGiHasVertexExpression
 - tolerance should be Double
```

PR #566, opened 2019-12-11
IAT-3514: Item Tabulator: Add Media Validation

bdrainer:
These are common changes needed for ItemTabulator.  I added a method to get attachments by section.  I needed this for media type in the ItemTabulator.  Media type validation only looks at Audio, ASL, and CC files.  Not images, braille files, etc.  

There will be a tabulator PR once this is merged and common is released.


src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -76,6 +77,58 @@
...
     }
 
+    /**
+     * Given a section a list of {@link DataStoreAttachment} are returned associated with that section.
+     *
+     * @param item    The item to look for attachments on.
+     * @param section The section to look for attachments.
+     * @return A list of data store attachments.
+     */
+    public static List<DataStoreAttachment> getDataStoreAttachmentsBySection(Item item, String section) {
```
millerts: If I remember correctly the TIMS model has empty attachment references created?  As in the JSON model contains properties that have 1-n attachments that are non-null even though the attachment doesn't really exist?  

I ask because I was going to suggest some refactoring and streamlining, but it breaks down if the item model starts with invalid attachment objects

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -5,7 +5,7 @@
 import org.apache.commons.collections4.CollectionUtils;
 import org.apache.commons.collections4.ListUtils;
 import org.apache.commons.io.FileUtils;
-import org.apache.commons.lang.StringUtils;
+import org.apache.commons.lang3.StringUtils;
```
malaffoon: I know we have a habit of sticking our heads in the sand regarding tech debt but do we have a ticket to upgrade to lang3 throughout?   @millerts

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -76,6 +77,58 @@
...
+    /**
+     * Given a section a list of {@link DataStoreAttachment} are returned associated with that section.
+     *
+     * @param item    The item to look for attachments on.
+     * @param section The section to look for attachments.
+     * @return A list of data store attachments.
+     */
+    public static List<DataStoreAttachment> getDataStoreAttachmentsBySection(Item item, String section) {
+        List<DataStoreAttachment> attachments = new ArrayList<>();
+        if (item != null) {
```
malaffoon: Be a bit more readable to use guard style code here, i.e. 
```
if (item == null) return attachments;
```

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
```diff
@@ -76,6 +77,58 @@
...
+                default:
+                    break;
+            }
+        }
+        return attachments.stream().filter(a -> StringUtils.isNotEmpty(a.getAttachment().getFileName()))
+            .collect(Collectors.toList());
+    }
+
+    /**
+     * Given a section a list of file names associated with that section.
```
malaffoon: nit: not quite proper english? perhaps `Given a section, return a list ...`

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
millerts: There isn't a ticket.  Brett or you can create one and put it into "Tech Debt".

malaffoon: I'll do that. Kind of like asking Santa Claus for World Peace, but what the heck.

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
bdrainer: The new method at the very end filters out attachments with empty file names.  This what we are doing in the existing method `getDataStoreAttachments`.  Let me know how I can streamline this and I'll make those changes.

src/main/java/org/opentestsystem/ap/common/model/attachment/AttachmentSelector.java
bdrainer: It's a good observation.  I noticed the ItemTabulator, and maybe I made it this way, doesn't have "commons.lang" but only "commons.lang3".  So I started noticing in other places (i.e. common) where we are using just "lang" and changed it to "lang3".  It would be nice to move everything to "3" so I'm in favor of a tech debt ticket for this.

bdrainer:
The suggested changes have been made.  Merging now.  I'll circle back if there are streamlined suggestions as mentioned in the PR.

PR #567, opened 2019-12-16
Feature/iat 3542

bdrainer:
The primary change is making the method "public" in the EbsrTransformer.  The TIMS scoring service uses the transformers to generate the QRX as a string. All, except Ebsr, was already public.

The other change is adding a configuration for the typical beans needed to generate SAAIF files and item.json.  Most services configure those beans in their respective app level configuration.  It seems like a good idea to have this.


bdrainer:
I'll merge now but feedback is welcome and I will apply the suggestions as soon as I can and get a new PR up with them.

src/main/java/org/opentestsystem/ap/common/config/SaaifConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    @Bean
+    @ConditionalOnMissingBean
+    public JsonModelAssembler jsonModelAssembler() {
+        log.info("Creating bean jsonModelAssembler");
+        return new JsonModelAssembler();
+    }
+
+    // ------------------------------------------------------------------------
```
millerts: What does this line represent?

src/main/java/org/opentestsystem/ap/common/config/SaaifConfiguration.java
```diff
@@ -0,0 +1,119 @@
...
+        this.restTemplateBuilder = restTemplateBuilder;
+        this.freemarker = freemarker;
+    }
+
+    // ------------------------------------------------------------------------
+
+    @Bean
+    @ConditionalOnMissingBean
+    public ImageGenerator imageGenerator() {
+        log.info("Creating bean imageGenerator");
```
millerts: Can we change all the logging in this class to debug?

src/main/java/org/opentestsystem/ap/common/config/SaaifConfiguration.java
bdrainer: I deleted these.  We spoken about these before.

src/main/java/org/opentestsystem/ap/common/config/SaaifConfiguration.java
bdrainer: I like info here.  This only happens on startup and I like seeing the beans as they are created.  Some of the other config classes are doing this like DatabaseConfiguration.

PR #568, opened 2019-12-17
IAT-3542: Refactored the ItemBankProperties.service.  It is more gene…

bdrainer:
…ric to allow for easier addition of new services.


bdrainer:
I'll merge now as I'm going to pin and release the scoring service to QA.  Any feedback here I'll address in the next PR.

src/main/java/org/opentestsystem/ap/common/config/ItemBankProperties.java
```diff
@@ -156,31 +156,47 @@
...
+            // the resource service are given defaults that are likely to never change,
+            // if they do change then updating the configuration is a way to override without having to change code
+            this.permission  = new ServiceResourceProps();
+            this.permission.setResourcePath("/auth/users/");
+
+            this.timsScoring = new ServiceResourceProps();
+            this.timsScoring.setResourcePath("/api/v1/score/");
+
+            this.tdsScoring = new ServiceResourceProps();
+            this.tdsScoring.setResourcePath("/itemscoring/Scoring/ItemScoring");
```
millerts: What are the rules for when one should add a service path to ServiceResourceProps?

bdrainer: This is kind of like defining a Zuul route.  The scenario is a service needs to call another service.  The API gateway needs to call permissions, for example, to get user's permission.

The service prop gives the URL to a service, the base URL.  If that is all that is needed then a ServiceProp is the right choice.

The ServiceResourceProp allows for the URL and a resource path.  Scoring  and permission have a single resource so there is only a single property for it.  If we need more than a single resource path we could change/enhance this.

QTI has a similar configuration concept.

```app:
  item-definition-client:
    base-url:  http://ap-item-translation-service
    pass-through-headers-filter:
      - Authorization```

millerts: Werd.  Sounds like IMRT may need to leverage this pattern rather than its current pattern because it makes calls to IMS and Validation.

bdrainer: What are your thoughts on putting these "clients" in the common "restclient" package.  There is a Feign one there now for the history service (HistoryClient) which IMS uses.  There could be others like an ItemManagementClient that talks to IMS.  It doesn't have to be Feign.  The PermissionServiceClient, which is not located in the "restclient" package but could be, is using RestTemplate.

PR #570, opened 2019-12-17
IAT-3545 - Added default values to IMRT fields

alex-ponce:


src/main/java/org/opentestsystem/ap/common/imrt/model/BaseItem.java
```diff
@@ -157,9 +157,9 @@
     private UUID commitId;
     private String itemVersion;
 
-    private String pencilPaper;
-    private String stateSpecificUse;
-    private String stateSpecificUseDescription;
+    private String pencilPaper = EMPTY;
```
millerts: 👍 - we need to move this to Lombok.  Or remove hibernate.  My vote is the latter :D

